PR #3, opened 2015-04-08
KYLIN-666 Select float type column got class cast exception

romansew:
SQL like:
select float_column from test_fact_table
got exception:
java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Float


liyang-gmt8:
This request can be closed.

romansew:
Thanks!

PR #5, opened 2015-05-21
KYLIN-752 improve query performance for values that are not in dictio…

superhua:
A quick fix to drop the not-in-dictionary values before generating hbase scan range.
And based on it, we can make a further optimization that as long as there is at least one andFilter is empty(after we drop the not-in-dictionary values), we can simply assert that the whole andFilters is empty to avoid an unnecessary hbase scan.
However, this strategy only works for 'EQ', 'IN'. For 'LT', 'LTE', 'GT', 'GTE', we have to replace them with the some sort of closest smaller or bigger values by calling "final public int getIdFromValue(T value, int roundingFlag)" of "Dictionary" class.

### DIFF
--- a/storage/src/main/java/org/apache/kylin/storage/hbase/CubeStorageEngine.java
+++ b/storage/src/main/java/org/apache/kylin/storage/hbase/CubeStorageEngine.java
@@ -421,8 +421,10 @@ public class CubeStorageEngine implements IStorageEngine {
             }
 
             Collection<ColumnValueRange> andRanges = translateToAndDimRanges(andFilter.getChildren(), cubeSegment);
-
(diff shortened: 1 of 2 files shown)


liyang-gmt8:
I'll merge this pull request in the coming week if no other interruptions. JIRA https://issues.apache.org/jira/browse/KYLIN-740

Sorry for the late response, an urgent PoC kept me very busy for last two weeks.

liyang-gmt8:
Merged into 0.7-staging eaf66856178feb79ebe55467808c476520da55c6
Followed by refactor aaec680182712ed4906d1548213688de92cc5538

Many thanks superhua!

PR #13, opened 2015-10-12
Fix for KYLIN-1070

vipinkumar7:
fact table is getting saves as it is without upper case


vipinkumar7:
fact table is getting saves as it is without upper case

PR #17, opened 2015-12-19
support specifying different kylin_job_conf.xml(isolation of hadoop job queues) for different kylin projects

IT-DONG:
1. for one hadoop cluster, we sometimes want to assign different hadoop job queues(machine resources) for different kylin projects(different business group), thus, adding logic to auto detect the kylin_job_conf_medium/small/large_projectname.xml first, then kylin_job_conf_projectname.xml, then kylin_job_conf_medium/small/large.xml, finally kylin_job_conf.xml, if still not found, throws an exception. It means we can support project level kylin job configuration file for different projects. and the piece of codes have been tested by my self with building cubes, refreshing cubes, and etc, which works successfully.

2. currently, commit 'ead10dc' is for the case that when users are in China mainland, we may be blocked with some download urls, therefore, we cannot build the front end just one time, it is possible for us to try many times to build the front end without building the back end again and again, so, for the script/build.sh, I added a special check, when running: sh build.sh fe, it builds the front end only, when running: sh build.sh be, it builds the back end only, when running: sh build.sh, it builds both the fe and be components.

### DIFF
--- a/atopcalcite/pom.xml
+++ b/atopcalcite/pom.xml
@@ -27,7 +27,7 @@
     <parent>
         <groupId>org.apache.kylin</groupId>
         <artifactId>kylin</artifactId>
-        <version>1.1.1-incubating-SNAPSHOT</version>
+        <version>1.1.1-incubating</version>
     </parent>
 
     <properties>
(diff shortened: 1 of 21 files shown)


lukehan:
Hi Dong,
    Could you please open JIRA and attache your patch there for review and merge?
    And, please submit to 1.x-staging for your patch, master branch is always latest stable release one.

    Thanks.
Luke

IT-DONG:
,move the pull request to 1.x-staging branch

PR #29, opened 2016-03-18
KYLIN-1504 Use NavigableSet to store rowkey and use prefix filter to check resource path prefix instead String comparison on tomcat side

haoch:
https://issues.apache.org/jira/browse/KYLIN-1504

- Use NavigableSet<String> instead ArrayList<String> to store natively ordered and unique row-key instead of ugly repeatedly using `Collections.sort` or check whether existing on business logic layer, in fact because the raw-key is originally sorted in hbase, the change won't consume any more computational complexity.
- Verify prefix in hbase region level using prefix filter instead of comparing String in client side.

### DIFF
--- a/core-common/src/main/java/org/apache/kylin/common/persistence/FileResourceStore.java
+++ b/core-common/src/main/java/org/apache/kylin/common/persistence/FileResourceStore.java
@@ -26,6 +26,8 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.NavigableSet;
+import java.util.TreeSet;
 
 import org.apache.commons.io.IOUtils;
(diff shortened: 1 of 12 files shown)


haoch:
liyang added a comment - 5 minutes ago - edited
Thanks Hao! Merged #29 with some revision. Didn't include the PrefixFilter commit because a) range scan is equally good performance[1]; b) want to keep the KeyOnlyFilter for reduced traffic.
[1] http://stackoverflow.com/questions/10942638/should-i-user-prefixfilter-or-rowkey-range-scan-in-hbase
Reply

haoch:
Close the pull request because the patch has been merged into https://github.com/apache/kylin/commit/801fb83b22e6a737ca9c43155a4860951bf370a2

PR #30, opened 2016-03-18
KYLIN-1505 Combine guava filters with Predicates.and

haoch:
https://issues.apache.org/jira/browse/KYLIN-1505

* Combine guava filters with Predicates.and(filters)
* Combine hbase RowKeyOnlyFilter and PrefixFilter with FilterLists


haoch:
Combine changes into #31, so that only need to merge single patch into master, so treat this one as resolved.

PR #39, opened 2016-10-14
hive job use overrided MR job configuration by cube properties

ljl1988com:
This patch allows hive mr job use overrided MR job configuration by cube properties.
The test case which uses hadoop-2.6.0,hbase-1.1.3 and hive-1.1.0 passes as expected.


ljl1988com:
@liyang-gmt8 @shaofengshi ,Can you help me review it?

PR #49, opened 2017-03-09
changed unsafe code to safety

wangyuxue:
changed unsafe code to safety


asfgit:

PR #60, opened 2017-06-01
support jdbc source

phillipcheng:
jdbc source support


asfgit:

PR #61, opened 2017-06-26
Fix for KYLIN-2668

joeswingle:
Added the unit test and url prefix code to allow jdbc:kylin:ssl=True to be used in URL, not just java properties


yiming187:
Hi @joeswingle , thanks for the patch. Could you generate a patch for this change? Kylin has patch and review process, not PR and merge process. To learn more: http://kylin.apache.org/development/howto_contribute.html

asfgit:

PR #62, opened 2017-08-25
fix DefaultClient may not release connection when sync wipe cache

lichaojacobs:
### fix DefaultClient may not release connection when sync wipe cache

here  is error log which I added to find why sync ofthen fail

```
Thread failed during wipe cache at java.lang.IllegalStateException: Invalid use of BasicClientConnManager: connection still allocated.

```
so I add some code to make sure connection to be released


yiming187:
Thank you for the PR. Could you follow the patch process, with Kylin code style? please check: http://kylin.apache.org/development/howto_contribute.html

PR #72, opened 2017-09-26
KYLIN-2645 Upgrade to Kafka 0.11.0.1.

shuiqing301:


yiming187:
please hold this upgrade. If a customer is still using the old version, this will break the compatibility. This is the same case we should not upgrade Hive dependency.

PR #73, opened 2017-09-26
 KYLIN-2614 Fix defaultX509TrustManager implemente default methods.

shuiqing301:


shaofengshi:
hello shuiqing301, thanks for the patch; as an Apache project, Kylin accepts the contributions with a JIRA and then attaching the git patch file to it. Thanks for joining the community! See: https://kylin.apache.org/development/howto_contribute.html

PR #79, opened 2017-10-20
Apache kylin 2723

kyotoYaho:


core-common/src/main/java/org/apache/kylin/common/KylinConfigBase.java
```diff
@@ -1315,8 +1315,56 @@ public String getPerfLoggerClassName() {
...
+    public boolean isKylinMetricsMonitorEnabled() {
+        return Boolean.parseBoolean(getOptional("kylin.core.metrics.monitor-enabled", "false"));
+    }
+
+    public boolean isKylinMetricsReporterForQueryEnabled() {
+        return Boolean.parseBoolean(getOptional("kylin.core.metrics.reporter-query-enabled", "false"));
+    }
+
+    public boolean isKylinMetricsReporterForJobEnabled() {
+        return Boolean.parseBoolean(getOptional("kylin.core.metrics.reporter-job-enabled", "true"));
```
lidongsjtu: Better to disable job metrics before system cube check-in?

core-common/src/main/java/org/apache/kylin/common/KylinConfigBase.java
```diff
@@ -1315,8 +1315,56 @@ public String getPerfLoggerClassName() {
...
                 "org.apache.kylin.metrics.lib.impl.StubReservoir");
     }
+
+    public String getKylinSystemCubeSinkDefaultClass() {
+        return getOptional("kylin.core.metrics.system-cube-sink-default-class",
+                "org.apache.kylin.metrics.lib.impl.hive.HiveSink");
+    }
+
+    public String getKylinMetricsSubjectSuffix() {
+        String suffix = getOptional("kylin.core.metric.subject-suffix", null);
```
lidongsjtu: How about writing as single-line:
getOptional("kylin.core.metric.subject-suffix", getDeployEnv());

core-job/src/main/java/org/apache/kylin/job/execution/ExecuteResult.java
```diff
@@ -30,6 +30,14 @@
 
     private final State state;
     private final String output;
+    private final Throwable throwable;
+
+    /**
+     * Default constructor to indicate a success ExecuteResult.
+     */
+    public ExecuteResult() {
```
lidongsjtu: Better to use static to create instance with state, and keep constructor generic, for example:
ExecuteResult.createSucceeded()
ExecuteResult.createError()
...

core-storage/src/main/java/org/apache/kylin/storage/gtrecord/GTCubeStorageQueryBase.java
```diff
@@ -272,6 +274,22 @@ private FunctionDesc findAggrFuncFromCubeDesc(FunctionDesc aggrFunc) {
         return resultD;
     }
 
+    private long getQueryFilterMask(Set<TblColRef> filterColumnD) {
+        long filterMask = 0;
+
+        logger.info("Filter column set for query: " + filterColumnD.toString());
```
lidongsjtu: Set.toString() will not return joined string by default, have a check?

lidongsjtu: And will filterColumnD be null? If yes, NPE might happen here.

engine-mr/src/main/java/org/apache/kylin/engine/mr/CubingJob.java
```diff
@@ -61,20 +64,50 @@
...
+                if (jobTypeEnum.name.equals(name.toUpperCase())) {
+                    return jobTypeEnum;
+                }
+            }
+            return null;
+        }
+    }
+
+    //32MB per block created by the first step
+    public static final long MIN_SOURCE_SIZE = 33554432L;
```
lidongsjtu: could be private?

server-base/src/main/java/org/apache/kylin/rest/response/SQLResponse.java
```diff
@@ -185,4 +204,26 @@ public boolean isStorageCacheUsed() {
...
+        }
+    }
+
+    public void setCubeSegmentStatisticsList(
+            List<QueryContext.CubeSegmentStatisticsResult> cubeSegmentStatisticsList) {
+        try {
+            this.queryStatistics = cubeSegmentStatisticsList == null ? null
+                    : SerializationUtils.serialize((Serializable) cubeSegmentStatisticsList);
+        } catch (Exception e) { // serialize exception should not block query
+            System.out.println("Error while serialize queryStatistics due to " + e);
```
lidongsjtu: we should avoid System.out.println() in production code

core-common/src/main/java/org/apache/kylin/common/KylinConfigBase.java
kyotoYaho: Agree

core-common/src/main/java/org/apache/kylin/common/KylinConfigBase.java
kyotoYaho: Good suggestion

core-job/src/main/java/org/apache/kylin/job/execution/ExecuteResult.java
kyotoYaho: Make Sense, will refine this part.

core-storage/src/main/java/org/apache/kylin/storage/gtrecord/GTCubeStorageQueryBase.java
kyotoYaho: HashSet.toString() will return like [,,...]

filterColumnD will not be null. Since it's initialized by 
Set<TblColRef> filterColumnD = Sets.newHashSet();

engine-mr/src/main/java/org/apache/kylin/engine/mr/CubingJob.java
kyotoYaho: changed

PR #92, opened 2018-01-06
KYLIN-2881+KYLIN-1872 review

lidongsjtu:


core-common/src/main/java/org/apache/kylin/common/QueryContext.java
```diff
@@ -138,6 +136,48 @@ public long addAndGetScannedBytes(long deltaBytes) {
...
+
+    /**
+     * stop the whole query by rest call
+     */
+    public void stopEarly(String reason) {
+        stopQuery(null, reason);
+    }
+
+    private void stopQuery(Throwable t, String reason) {
+        if (isStopped()) {
```
binmahone: 这里应该完整的使用compareandset，否则可能两个线程都会进入并调用listender的stop方法

kyotoYaho: Since isRunning is AtomicBoolean, I think it's thread-safe. The first running into this part will short-circuit others.

kyotoYaho:
IT 可以有>_<

core-common/src/main/java/org/apache/kylin/common/QueryContext.java
kyotoYaho: You are right. It's better to use compareAndSet.

kyotoYaho:
Interrupt 的功能已经被其它的exception所覆盖了，例如KylinTimeoutException。这里的interrupt的逻辑是在sql中如果有任何exception，就会interrupt掉当前的query

PR #105, opened 2018-02-24
KYLIN-2884 Add delete segment function for portal - bug fix

sanjulian:


asfgit:

PR #107, opened 2018-02-27
kylin-3119

pengjianhua:


asfgit:

query/src/main/java/org/apache/kylin/query/util/QueryUtil.java
```diff
@@ -50,32 +50,39 @@ public static String massageSql(String sql, String project, int limit, int offse
...
+            if (offset > 0 && matcher.group(2) == null) {
+                sql1 += ("\nOFFSET " + offset);
+            }
         }
 
         // https://issues.apache.org/jira/browse/KYLIN-2649
-        if (kylinConfig.getForceLimit() > 0 && !sql.toLowerCase().contains("limit")
-                && sql.toLowerCase().matches("^select\\s+\\*\\p{all}*")) {
-            sql += ("\nLIMIT " + kylinConfig.getForceLimit());
+        if (kylinConfig.getForceLimit() > 0 && !sql1.toLowerCase().contains("limit")
```
rogercloud: Here's another place that check "limit", would you please consider it in this fix?

A possible fix is record the previous match result, telling whether the SQL contains limit or not.

query/src/test/java/org/apache/kylin/query/util/QueryUtilTest.java
```diff
@@ -51,6 +51,58 @@ public void testMassageSql() {
...
+            String sql = "select count(*) test_limit_and_offset from test_kylin_fact where price > 10.0";
+            String s = QueryUtil.massageSql(sql, "default", 50000, 50, "DEFAULT");
+            Assert.assertEquals(
+                    "select count(*) test_limit_and_offset from test_kylin_fact where price > 10.0\n" +
+                            "LIMIT 50000\nOFFSET 50",
+                    s);
+        }
+
+        {
+            String newLine = System.getProperty("line.separator");
```
rogercloud: I guess you want to test newline character on different platform. It's a good point. I suggest to hard code "\r\n" for Windows and "\n" for Linux instead of reading "line.separator". Maybe I misunderstand, please figure it out.

As there are more code blocks here, some comments to explain the test purpose is a good practice. :)

pengjianhua:
hi,rogercloud.the  `kylinConfig.getForceLimit()` add a check too.
And  the reason that I use "line.separator" is in function massageSql has the code 
`sql = sql.replace("\r", " ").replace("\n", System.getProperty("line.separator"));`
In windows,if I use '\n',it will be replaced by '\r\n',
if I use '\r\n',it will be replaced by ' \r\n'

PR #108, opened 2018-02-27
Apache kylin 2895

kyotoYaho:


asfgit:

shaofengshi:
When enabling this feature, Kylin build mapper got OOM in the "Build cube inmem step". We think there is some room to improve.

PR #116, opened 2018-03-16
KYLIN-3292

xingpeng1:
KYLIN-3292 The update button should be disabled with none key and value in setting config dialog


asfgit:

PR #117, opened 2018-03-16
KYLIN-3293, fix FixedLenHexDimEnc.java return a wrong code length lea…

tttMelody:
…ds to cut bytes error.


asfgit:

PR #130, opened 2018-04-18
Analytic table name ArrayIndexOutOfBoundsException

PingGao89:
Called without a default parameter will result in
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: 2
	at org.apache.kylin.storage.hbase.util.DeployCoprocessorCLI.main(DeployCoprocessorCLI.java:118)
In 100 lines to determine if the first parameter of the command line is default, curIdx has ++, if the first parameter is not the default line 103, curIdx is re-++, resulting in the parameter misalignment.


asfgit:

shaofengshi:
You're correct, thanks Gao Ping!

PR #135, opened 2018-04-25
KYLIN-3347 : calcite function QueryService Exception

imsong:
sorry. Request again.

### DIFF
--- a/query/src/main/java/org/apache/kylin/query/util/DefaultQueryTransformer.java
+++ b/query/src/main/java/org/apache/kylin/query/util/DefaultQueryTransformer.java
@@ -43,15 +43,13 @@ public class DefaultQueryTransformer implements IQueryTransformer {
     private static final Pattern PTN_INTERVAL = Pattern.compile(
             "interval" + SM + "(floor\\()([\\d\\.]+)(\\))" + SM + "(second|minute|hour|day|month|year)",
             Pattern.CASE_INSENSITIVE);
-    private static final Pattern PTN_HAVING_ESCAPE_FUNCTION = Pattern.compile("\\{fn" + SM + "(EXTRACT\\(.*?\\))" + "\\}",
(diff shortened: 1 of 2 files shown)


asfgit:

rogercloud:
@imsong Some UT failes, please have a check.

query/src/main/java/org/apache/kylin/query/util/DefaultQueryTransformer.java
```diff
@@ -50,6 +50,8 @@
     private static final Pattern PIN_SUM_OF_FN_CONVERT = Pattern
             .compile(S0 + "SUM" + S0 + "\\(" + S0 + "\\{\\s*fn" + SM + "convert" + S0 + "\\(" + S0 + "([^\\s,]+)" + S0
                     + "," + S0 + "SQL_DOUBLE" + S0 + "\\)" + S0 + "\\}" + S0 + "\\)", Pattern.CASE_INSENSITIVE);
+    private static final Pattern PTN_HAVING_FUNCTION = Pattern.compile("\\{fn" + "(.*?)" + "\\}",
```
rogercloud: It's too general as it matches all functions. How about adding a CURRENT_TIMESTAMP function like PTN_MIN_1 instead?

query/src/main/java/org/apache/kylin/query/util/DefaultQueryTransformer.java
```diff
@@ -138,6 +140,12 @@ public String transform(String sql, String project, String defaultSchema) {
             sql = sql.substring(0, m.start(1)) + "'" + value + "'" + sql.substring(m.end(3));
         }
 
+        while (true) {
```
rogercloud: Please add some comments to explain how the pattern is used.

query/src/main/java/org/apache/kylin/query/util/DefaultQueryTransformer.java
imsong: I also need to use TIMESTAMPADD and SQL_TSI_HOUR functions.
Because, Tableau + kylin odbc generate automatically query using this functions.

I don't think that is great idea... I hope that Kylin can support calcite reserved functions. https://calcite.apache.org/docs/reference.html

thank you.

shaofengshi:
LGTM, thanks for the PR!

PR #136, opened 2018-04-25
KYLIN-3349 : Spark Numberformat Exception solved

imsong:


asfgit:

shaofengshi:
As we discussed in the JIRA, the root cause is a wrong measure definition be generated, this PR  will not be accepted. Thanks for the reporting!

PR #146, opened 2018-05-29
KYLIN-3299, fix when year less than 1000 in int data type and date encoding

Aaaaaaron:
…code.


asfgit:

PR #147, opened 2018-06-03
KYLIN-3161 Enforce global dictionary for bitmap count distinct column…

Wayne1c:
…(as data type is not int)

### DIFF
--- a/core-cube/src/main/java/org/apache/kylin/cube/model/validation/rule/DictionaryRule.java
+++ b/core-cube/src/main/java/org/apache/kylin/cube/model/validation/rule/DictionaryRule.java
@@ -19,8 +19,10 @@
 package org.apache.kylin.cube.model.validation.rule;
 
 import java.util.ArrayList;
+import java.util.HashMap;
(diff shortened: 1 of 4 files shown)


asfgit:

core-cube/src/main/java/org/apache/kylin/cube/model/validation/rule/DictionaryRule.java
```diff
@@ -89,17 +100,32 @@ public void validate(CubeDesc cubeDesc, ValidateContext context) {
                 return;
             }
 
+            if (StringUtils.isNotEmpty(builderClass) && !builderClass.equalsIgnoreCase(GlobalDictionaryBuilder.class.getName()) && bitmapMeasures.containsKey(dictCol) && !dictCol.getColumnDesc().getType().isIntegerFamily()){
```
shaofengshi: "builderClass" can be empty, which would means the normal dictionary. So this case should also be checked.

shaofengshi:
LGTM

PR #153, opened 2018-06-11
KYLIN-1948, add test to cover the case.

tttMelody:


asfgit:

PR #154, opened 2018-06-12
KYLIN-3388 remove create redistribute flat hive table step if not set…

Wayne1c:
… distribute column


asfgit:

PR #156, opened 2018-06-20
KYLIN-3399, makes the code that added in KYLIN-2474 more intuitive.

tttMelody:


asfgit:

PR #159, opened 2018-06-27
KYLIN-3403 filter code system add one type of Date

Wayne1c:


asfgit:

core-metadata/src/main/java/org/apache/kylin/metadata/filter/FilterCodeSystemFactory.java
```diff
@@ -95,4 +99,32 @@ public int compare(Object o, Object t1) {
...
+        }
+
+        @Override
+        public int compare(String o1, String o2) {
+            try {
+                long d1 = DateFormat.stringToMillis(o1);
+                long d2 = DateFormat.stringToMillis(o2);
+                return Long.compare(d1, d2);
+            } catch (IllegalArgumentException e) {
+                return o1.compareTo(o2);
```
shaofengshi: The exception shouldn't be caught silently; It should throw an error here, as using string comparator may return the wrong result.

shaofengshi:
Yifan has fixed this; The pr is not needed;

PR #160, opened 2018-07-02
KYLIN-3434 Support prepare statement in Kylin server side

allenma:
support cache preparedStatement in kylin sever, to save sql parse/optimize/implement time. Detail information can see in: https://issues.apache.org/jira/browse/KYLIN-3434


asfgit:

PR #161, opened 2018-07-04
KYLIN-3440 Couldn't merge if there is an empty segment

luguosheng1314:


asfgit:

PR #164, opened 2018-07-05
KYLIN-3439, make default kylin.web.timezone in KylinConfigBase the same with kylin-defaults.properties

tttMelody:
…me with kylin-defaults.properties


asfgit:

PR #168, opened 2018-07-18
Kylin 3255 cannot save cube

GinaZhai:
modify the code for null condition


asfgit:

PR #170, opened 2018-07-19
KYLIN-3457 use cube level config

Wayne1c:


asfgit:

PR #171, opened 2018-07-23
Superset document

GinaZhai:
including Chinese and English document


asfgit:

PR #172, opened 2018-07-24
KYLIN-3453 Improve cube size estimation for topn,count distinct

Wayne1c:


asfgit:

PR #183, opened 2018-08-01
KYLIN-3461 metastore.sh refresh-cube-signature not updating cube sign…

GinaZhai:


asfgit:

PR #190, opened 2018-08-07
v2.4.0 Convert Cuboid Data to HFile configuration

momoxixi:
When Convert Cuboid Data to HFile, kylin `org.apache.kylin.storage.hbase.steps.CubeHFileJob` not add `org.apache.kylin.engine.mr.common.AbstractHadoopJob` conf to mapreduce.
So when change the `$KYLIN_HOME/conf/kylin_job_conf.xml`, it will not work actually.
Besides, there is a solution, to change the `kylin.properties`  that add `kylin.storage.url=default@hbase,key=value`. I don't konw if this is a kylin's feature.


shaofengshi:
Invalid PR, please avoiding create PR from another trunk (for example 2.4.x) to master.
The issue you mentioned is https://issues.apache.org/jira/browse/KYLIN-3438; The patch is attached in the JIRA, please manually apply it.

PR #198, opened 2018-08-14
KYLIN-3499 Update HBaseResourceStore.java

momoxixi:
Avoid resource inconsistent caused by hbase rpc timeout

### DIFF
--- a/storage-hbase/src/main/java/org/apache/kylin/storage/hbase/HBaseResourceStore.java
+++ b/storage-hbase/src/main/java/org/apache/kylin/storage/hbase/HBaseResourceStore.java
@@ -44,6 +44,7 @@ import org.apache.hadoop.hbase.client.Result;
 import org.apache.hadoop.hbase.client.ResultScanner;
 import org.apache.hadoop.hbase.client.Scan;
 import org.apache.hadoop.hbase.client.Table;
+import org.apache.hadoop.hbase.client.RetriesExhaustedException;
 import org.apache.hadoop.hbase.filter.CompareFilter;
 import org.apache.hadoop.hbase.filter.Filter;
 import org.apache.hadoop.hbase.filter.FilterList;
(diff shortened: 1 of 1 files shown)


asfgit:

shaofengshi:
@momoxixi hello, thanks for the reporting; Could you please create a JIRA for the problem, and then update the commit message as "KYLIN-XXXX <description>". Thanks!

momoxixi:
@shaofengshi  OK, I have been add a JIRA, please check it, thank you.

momoxixi:
In my Hadoop cluster, HDFS would occur `SLOW BlockReceiver xxx` sometimes, it  also made HBase rpc slow. When HBase rpc timeout kylin rollback the resource timestamp but actually HBase rpc already finished at server. So I added an another check to make sure HBase server has not received that rpc.

shaofengshi:
Hi Momoxixi,

Sorry for the late response. So you mean, the invocation of "table.checkAndPut" may throw a "RetriesExhaustedException" but actually, the change has been taken effectively in the HBase server side, right?

If so, two comments here:
1) This change adds a "try-catch" embedded in the original "try-catch", the embedded try-catch should be avoided.
2) Besides the "RetriesExhaustedException", are there other similar cases? Can it be generalized to "IOException"?

Thanks for your contribution to Kylin!

momoxixi:
Hi shaofengshi,
You get my point, and here is my opinion for two comments,:
1. I consider Embedded `try-catch` is a compromise plan, or it would modify more code.
2. `IOException` maybe a more smarter choice, but after reading hbase source code `RetriesExhaustedException` is the one exception that need to recheck.

shaofengshi:
This change adds one more try when got "RetriesExhaustedException", it couldn't avoid such issue completely, but it can reduce the possibility I think. So I'd like to accept it. Thank you momo.

PR #204, opened 2018-08-21
KYLIN-3442 Fact distinct columns in Spark

Wayne1c:


asfgit:

engine-spark/pom.xml
```diff
@@ -100,6 +100,50 @@
             <version>0.9.10</version>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>org.scala-lang</groupId>
+            <artifactId>scala-library</artifactId>
+            <version>2.11.0</version>
```
shaofengshi: It is better to put the scala version to the root pom.xml

engine-spark/src/main/java/org/apache/kylin/engine/spark/SparkBatchCubingJobBuilder2.java
```diff
@@ -61,7 +61,7 @@ public CubingJob build() {
         inputSide.addStepPhase1_CreateFlatTable(result);
 
         // Phase 2: Build Dictionary
-        result.addTask(createFactDistinctColumnsStep(jobId));
+        result.addTask(createFactDistinctColumnsStepSPARK(jobId));
```
shaofengshi: Please use "Spark" to replace "SPARK"

PR #207, opened 2018-08-22
KYLIN-3507 Avoid NPE when project is not found

ian4hu:
Fix [KYLIN-3507](https://issues.apache.org/jira/projects/KYLIN/issues/KYLIN-3507)


asfgit:

PR #212, opened 2018-08-27
KYLIN-3514 Support AUC UDAF

r7raul1984:


asfgit:

core-metadata/pom.xml
```diff
@@ -24,6 +24,18 @@
     <packaging>jar</packaging>
     <name>Apache Kylin - Core Metadata</name>
     <description>Apache Kylin - Core Metadata</description>
+    <build>
```
yiming187: Do you mean this commit only works in JDK8+?

r7raul1984: I use stream api in my code. And I found UT on master branch with JDK 1.7(deprecated): Kylin-Master-JDK-1.7  in kylin docs.So I add this config.

core-metadata/pom.xml
```diff
@@ -87,6 +99,10 @@
             <groupId>com.tdunning</groupId>
             <artifactId>t-digest</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.github.haifengl</groupId>
+            <artifactId>smile-core</artifactId>
```
shaofengshi: Could you please add the copyright of smile-core to Kylin's LICENSE file in the root folder?

shaofengshi:
How to use this function in Kylin? I didn't see the UDF implementation. Please provide some description and sample for us to understand this function. Thanks!

shaofengshi:
This is not a complete implementation, for example, it doesn't translate the query to this measure type. Will close it now, and revisit it later if find a total solution. Thank you!

PR #215, opened 2018-08-30
KYLIN-3518 Coprocessor reports NPE when execute a query on HBase 2.0

caolijun1166:


shaofengshi:
Merged with the patch, github's rebase doesn't work.

PR #218, opened 2018-09-03
minor, fix grunt script - font build

Emiya0306:


asfgit:

PR #223, opened 2018-09-07
KYLIN-3258 No check for duplicate cube name when creating a hybrid cube

GinaZhai:
Fix 2.5.0 bug


asfgit:

PR #225, opened 2018-09-10
KYLIN-3517 Upadate coprocessor on HBase2.0 is avaliable.

caolijun1166:
KYLIN-3517 Upadate coprocessor on HBase2.0 is avaliable.


asfgit:

PR #230, opened 2018-09-12
KYLIN-3525 Reserve intermediate flat table data if kylin.source.hive.keep-flat-table set to true

hit-lacus:
Reserve intermediate flat table data if kylin.source.hive.keep-flat-table set to true


asfgit:

PR #241, opened 2018-09-21
KYLIN-3546 Add column which belongs to root fact table in join relation but lost

hit-lacus:


source-kafka/src/main/java/org/apache/kylin/source/kafka/KafkaInputBase.java
```diff
@@ -89,7 +95,28 @@ public DataModelDesc getDataModel() {
...
+                }
+                // Add column which belongs to root fact table in join relation but lost
+                for (JoinTableDesc joinTableDesc : getDataModel().getJoinTables()) {
+                    JoinDesc jd = joinTableDesc.getJoin();
+                    for (TblColRef colRef : jd.getForeignKeyColumns()) {
+                        if (colRef.getTableRef().equals(rootFactTable)) {
+                            factTableColumnSet.add(colRef);
+                        }
+                    }
+                    for (TblColRef colRef : jd.getPrimaryKeyColumns()) {
```
shaofengshi: PK columns are on lookup tables, they won't be fact table, so this loop has no necessary to be here.

hit-lacus: OK, I will fix it.

PR #245, opened 2018-09-25
KYLIN-3586

caolijun1166:
Solve KYLIN-3586: Use parseInt() to parse a primitive instead of to use boxing/unboxing


asfgit:

asfgit:

PR #254, opened 2018-09-25
KYLIN-3562, refine the logic that updating the user when user logged in.

tttMelody:


asfgit:

asfgit:

shaofengshi:
Hi jiatao, as the root cause is KYLIN-3499, this change is not needed I think. Let's see how it goes on. Thank you.

PR #281, opened 2018-10-08
KYLIN-2924 disable google error-prone for travis-ci

shaofengshi:
Travis-ci failed with message:
"
The job exceeded the maximum log length, and has been terminated."

I try to disable errorprone check to reduce the log output.


shaofengshi:
Just try if it can make travis-ci happy

PR #286, opened 2018-10-12
KYLIN-3617 Use job's cache in job scheduler

nichunen:


asfgit:

core-job/src/main/java/org/apache/kylin/job/dao/ExecutableDao.java
```diff
@@ -391,4 +405,13 @@ public void deleteJobOutput(String uuid) throws PersistentException {
             throw new PersistentException(e);
         }
     }
+
+    public void reloadAll() throws IOException {
+        try (AutoReadWriteLock.AutoLock lock = executableDigestMapLock.lockForWrite()) {
+            executableDigestCrud.reloadAll();
```
shaofengshi: Why this is in "lockForWrite()" but the next reload is in "lockForRead"?

nichunen: @shaofengshi You are right, it's a mistake, should use lockForWrite during reloading. PR updated, thanks

PR #290, opened 2018-10-15
KYLIN-3632 Add configuration that can switch on/off preparedStatement cache

allenma:
Add configuration that can switch on/off preparedStatement cache in kylin server

### DIFF
--- a/core-common/src/main/java/org/apache/kylin/common/KylinConfigBase.java
+++ b/core-common/src/main/java/org/apache/kylin/common/KylinConfigBase.java
@@ -1457,6 +1457,10 @@ abstract public class KylinConfigBase implements Serializable {
         return Integer.parseInt(this.getOptional("kylin.query.statement-cache-max-num-per-key", String.valueOf(50)));
     }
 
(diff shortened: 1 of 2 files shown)


asfgit:

core-common/src/main/java/org/apache/kylin/common/KylinConfigBase.java
```diff
@@ -1457,6 +1457,10 @@ public int getQueryMaxCacheStatementInstancePerKey() {
         return Integer.parseInt(this.getOptional("kylin.query.statement-cache-max-num-per-key", String.valueOf(50)));
     }
 
+    public boolean isQueryPreparedStatementCacheEnable() {
+        return Boolean.parseBoolean(this.getOptional("kylin.query.statement-cache-enabled", "false"));
```
shaofengshi: Gang, can we enable it by default? It is a good feature.

allenma: Sure

PR #292, opened 2018-10-17
KYLIN-3634 when filter column has null value may cause incorrect query result

wangbo:


shaofengshi:
Thank you bo, we will run a full integration test for this patch.

PR #295, opened 2018-10-18
KYILN-3597 Use SecureRandom instead of Random

caolijun1166:
KYILN-3597 Use SecureRandom instead of Random


asfgit:

PR #298, opened 2018-10-19
KYLIN-3619 Some job won't clean up temp directory after finished

shaofengshi:


asfgit:

PR #300, opened 2018-10-21
KYLIN-3640 Fix cube with desc broken state can not be deleted

nichunen:


asfgit:

PR #301, opened 2018-10-21
KYLIN-3633 Avoid potential dead lock when building global dictionary

nichunen:


asfgit:

core-dictionary/src/main/java/org/apache/kylin/dict/DictionaryGenerator.java
```diff
@@ -75,13 +75,19 @@ public static IDictionaryBuilder newDictionaryBuilder(DataType dataType) {
...
 
-            boolean accept = builder.addValue(value);
+                boolean accept = builder.addValue(value);
 
-            if (accept && samples.size() < nSamples && samples.contains(value) == false)
-                samples.add(value);
+                if (accept && samples.size() < nSamples && samples.contains(value) == false)
+                    samples.add(value);
+            }
+        } catch (Throwable e) {
```
shaofengshi: Use "Throwable" is not recommended; please use a concrete exception type instead.

PR #302, opened 2018-10-21
KYLIN-3620 resolve "--" is treated as comment marker between singe qu…

caolijun1166:
KYLIN-3620 resolve "--" is treated as comment marker between singe quotes

### DIFF
--- a/query/src/main/java/org/apache/kylin/query/util/QueryUtil.java
+++ b/query/src/main/java/org/apache/kylin/query/util/QueryUtil.java
@@ -184,7 +184,7 @@ public class QueryUtil {
 
     public static String removeCommentInSql(String sql1) {
         // match two patterns, one is "-- comment", the other is "/* comment */"
-        final String[] commentPatterns = new String[] { "--[^\r\n]*", "/\\*[\\s\\S]*?\\*/" };
+        final String[] commentPatterns = new String[]{"--(?!.*\\*/).*?[\r\n]", "/\\*(.|\r|\n)*?\\*/"};
 
         for (int i = 0; i < commentPatterns.length; i++) {
             sql1 = sql1.replaceAll(commentPatterns[i], "");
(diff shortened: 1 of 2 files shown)


asfgit:

query/src/main/java/org/apache/kylin/query/util/QueryUtil.java
```diff
@@ -184,7 +184,7 @@ public static boolean isSelectStatement(String sql) {
 
     public static String removeCommentInSql(String sql1) {
         // match two patterns, one is "-- comment", the other is "/* comment */"
-        final String[] commentPatterns = new String[] { "--[^\r\n]*", "/\\*[\\s\\S]*?\\*/" };
+        final String[] commentPatterns = new String[]{"--(?!.*\\*/).*?[\r\n]", "/\\*(.|\r|\n)*?\\*/"};
```
shaofengshi: Is the code formatted?

query/src/test/java/org/apache/kylin/query/util/QueryUtilTest.java
```diff
@@ -223,9 +223,22 @@ public void testRemoveCommentInSql() {
...
+
+        String content = "        --  One-line comment and /**range\n" +
+                "/*\n" +
+                "Multi-line comment\r\n" +
+                "--  Multi-line comment*/\n" +
+                "select price as " +
+                "/*\n" +
+                "Multi-line comment\r\n" +
+                "--  Multi-line comment*/\n" +
+                "revenue from /*One-line comment-- One-line comment*/ v_lineitem;";
```
shaofengshi: I didn't see the test case covers the situation of "--" in a quoted string.

query/src/main/java/org/apache/kylin/query/util/QueryUtil.java
caolijun1166: Yes, it had been formatted.

PR #310, opened 2018-10-24
KYLIN-3646 Add instruction about rowkey sequence in "Advanced setting" page 

Emiya0306:


asfgit:

PR #315, opened 2018-10-25
KYLIN-3648 Use Path as the input of addResource

Wayne1c:


asfgit:

PR #316, opened 2018-10-26
KYLIN-3635, make a stronger constraint about implementing method rese…

tttMelody:
…t().


asfgit:

PR #319, opened 2018-10-29
KYLIN-3556, replace String.intern to avoid unexpected locking collisions

etherge:
Locking on an interned string can cause unexpected locking collisions with other part of code. A better option is **_Interner_** in guava.


asfgit:

PR #325, opened 2018-11-02
KYLIN-3663 fix deleting project error when project has more than one …

woyumen4597:
…table

### DIFF
--- a/server-base/src/main/java/org/apache/kylin/rest/service/ProjectService.java
+++ b/server-base/src/main/java/org/apache/kylin/rest/service/ProjectService.java
@@ -28,6 +28,7 @@ import java.util.Set;
 
 import javax.annotation.Nullable;
 
+import com.google.common.collect.Sets;
 import org.apache.directory.api.util.Strings;
 import org.apache.kylin.metadata.project.ProjectInstance;
 import org.apache.kylin.metadata.realization.RealizationType;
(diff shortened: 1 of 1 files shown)


asfgit:

woyumen4597:
retest this please

server-base/src/main/java/org/apache/kylin/rest/service/ProjectService.java
```diff
@@ -137,8 +138,10 @@ public ProjectInstance updateProject(ProjectInstance newProject, ProjectInstance
     @PreAuthorize(Constant.ACCESS_HAS_ROLE_ADMIN)
     public void deleteProject(String projectName, ProjectInstance project) throws IOException {
         Set<String> tables = project.getTables();
-        for (String table : tables) {
+        for (String table : Sets.newTreeSet(tables)) {
```
shaofengshi: What's the reason to create a new set? Can you provide more information?

woyumen4597: Since the unloading table operation will modify the set tables,while tables is traversed at the same time,which will cause concurrentException.

shaofengshi: I see, thank you!

PR #326, opened 2018-11-03
KYLIN-3597, fix the critical bug reported by sonar: Make sure this ex…

Aaaaaaron:
…pression can't be zero before doing this division.


asfgit:

Aaaaaaron:
Hi @shaofengshi, I do a simple modify just change the init value, could you assess if this change is correct?

Aaaaaaron:
When "totalWeight" is 0, means there are no nodes, so "thisWeight" is 0 too.

PR #328, opened 2018-11-07
 Refactor config chapter&deploy in cluster and add one tutorial about how to use mysql as metastore

DingqianZhou:
…how to use mysql as metastore

`jekyll s` [done]


asfgit:

PR #340, opened 2018-11-08
 add "use mysql as metastore" in cn&en

DingqianZhou:


asfgit:

PR #344, opened 2018-11-14
#KYLIN-3684, HIVE_LIB is not set or not resolved correctly

ZhengshuaiPENG:
Enhancement script for find-hive-dependency.sh


asfgit:

PR #345, opened 2018-11-15
KYLIN-3656  Improve HLLCounter performance

hit-lacus:
The current HLLCounter implementation has some room to improve performance, as we find in our product environment. Improvement related to getCountEstimate of HLLCounter and constructor of HLLCounter.

- Create HLLCounter from another HLLCounter, we can copy register(using System.arraycopy) instead of merge. (Constructor of HLLCounter)
- Precompute harmonic mean in the HLLCSnapshot to avoid doing this on the fly. (getCountEstimate of HLLCounter)

UnitTest has add cost duration compare.


asfgit:

hit-lacus:
Local CI pass.

PR #347, opened 2018-11-15
KYLIN-3666 HDFS metadata url not be recognized

Wayne1c:


asfgit:

engine-mr/src/main/java/org/apache/kylin/engine/mr/common/AbstractHadoopJob.java
```diff
@@ -494,6 +494,17 @@ public static KylinConfig loadKylinConfigFromHdfs(SerializableConfiguration conf
...
+        
+        // This is a bad example where the thread local KylinConfig cannot be auto-closed due to 
+        // limitation of MR API. It works because MR task runs its own process. Do not copy.
+        @SuppressWarnings("unused")
+        SetAndUnsetThreadLocalConfig shouldAutoClose = KylinConfig.setAndUnsetThreadLocalConfig(config);
+
+        return config;
+    }
+
+    public static KylinConfig loadKylinConfigFromHdfsNotThreadLocal(String uri) {
```
shaofengshi: The method name is confusing. I think the root cause is the bad sample in AbstractHadoopJob.loadKylinConfigFromHdfs(), we should fix that.

PR #354, opened 2018-11-20
KYLIN-3697 minor, only kylin.sh start check-env

DingqianZhou:


asfgit:

PR #355, opened 2018-11-21
detailed configurations could be override in the Cube-level

DingqianZhou:


asfgit:

shaofengshi:
Merged with patch, thank you Dingqian!

PR #356, opened 2018-11-22
KYLIN-3665 Partition time column may never be used

Wayne1c:


asfgit:

PR #357, opened 2018-11-23
KYLIN-3694 Bug fix and revert guava version

hit-lacus:


asfgit:

PR #358, opened 2018-11-23
KYLIN-3559 Use Splitter for splitting String

whuwb:
Issue : https://issues.apache.org/jira/browse/KYLIN-3559
Solution : 
1. Leverage Splitter in Guava to create split/splitByComma and refactor splitAndTrim method in StringUtil
2. Replace the string.split with StringUtil.split


asfgit:

PR #360, opened 2018-11-26
KYLIN-3665 Partition time column may never be added

FrancisLiang:
Fixed [KYLIN-3665]: Partition time column may never be added.


asfgit:

FrancisLiang:
> Hi Francis, the change looks good. While the commit message didn't comply with Apache's style. Could you please update the commit message to "KYLIN-3665 Partition time column may never be added" (start with JIRA ID, no [] needed). Thank you!

Just did it. Thanks for the information!

PR #361, opened 2018-11-26
KYLIN-3704 upgrade calcite version to 1.16.0-kylin-r2 version

vaful:


asfgit:

PR #366, opened 2018-11-28
KYLIN-3707 add configuration for setting isolation-level for sqoop

woyumen4597:


asfgit:

woyumen4597:
@shaofengshi Good suggestion,but I think each datasource has its own configuration,including adaptor and xml file,This sqoop configuration is like other config such as sql.case-sensitive is special for one datasource.So in my opinion  there is no need to add extra common configuration..

woyumen4597:
CI has passed.
image

PR #368, opened 2018-11-30
KYLIN-3187 Forbid calls to JDK APIs that use the default locale

hit-lacus:
Fix encode bugs.


asfgit:

hit-lacus:
```sql
select uid, act_type, count(*)  as cnt
from useraction as ua 
where act_type > '播放'
group by uid, act_type
order by uid, act_type;

```

This sql query result is correct compare Kylin with Hive.

PR #373, opened 2018-12-06
KYLIN-3406 Ignore execute output file lose

hit-lacus:
## Root Cause
When any single file of Kylin's metadata miss in HDFS, the "Monitor" Page will failed to display all job info and kylin process will throw FileNotFoundException as above. The page has a warning message.
## Fix solution
The better way should be : miss file should only affect specific task. And if only execute output miss, task should be success, otherwise that should be a fatal exeception. You have to fix it manually.

Test at https://issues.apache.org/jira/projects/KYLIN/issues/KYLIN-3406

### DIFF
--- a/core-job/src/main/java/org/apache/kylin/job/dao/ExecutableDao.java
+++ b/core-job/src/main/java/org/apache/kylin/job/dao/ExecutableDao.java
@@ -18,6 +18,7 @@
 
 package org.apache.kylin.job.dao;
 
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


asfgit:

hit-lacus:
Local CI pass.

hit-lacus:
@shaofengshi OK, I knew it.

hit-lacus:
@shaofengshi I have rewrite this patch, and local ci test pass, please review again.

core-job/src/main/java/org/apache/kylin/job/dao/ExecutableDao.java
```diff
@@ -393,7 +395,14 @@ public ExecutableOutputPO getJobOutput(String uuid) throws PersistentException {
             return result;
         } catch (IOException e) {
             logger.error("error get job output id:" + uuid, e);
-            throw new PersistentException(e);
+            if(e.getCause() instanceof FileNotFoundException) {
+                result = new ExecutableOutputPO();
+                result.setUuid(uuid);
+                result.setStatus(ExecutableState.SUCCEED.name());
+                return result;
+            }else {
```
shaofengshi: This seems not formatted; Please double check and ensure your IDE has been configured.

hit-lacus: OK, I have fixed it.

PR #376, opened 2018-12-09
KYLIN-3617 Add javadoc for magic number in isTaskExecutableOutput

hit-lacus:
The method `isTaskExecutableOutput` is quite confusing to me, so I add my understanding as javadoc to that method, and I guess it might be helpful to other developers.


asfgit:

PR #381, opened 2018-12-10
KYLIN-3715 ProjectL2Cache don't be invalidated when adding override config in cu…

walhl:
…be level

### DIFF
--- a/core-cube/src/main/java/org/apache/kylin/cube/CubeManager.java
+++ b/core-cube/src/main/java/org/apache/kylin/cube/CubeManager.java
@@ -151,11 +151,13 @@ public class CubeManager implements IRealizationProvider {
 
         @Override
         public void onProjectSchemaChange(Broadcaster broadcaster, String project) throws IOException {
-            for (IRealization real : ProjectManager.getInstance(config).listAllRealizations(project)) {
+            ProjectManager projectManager = ProjectManager.getInstance(config);
+            for (IRealization real : projectManager.listAllRealizations(project)) {
                 if (real instanceof CubeInstance) {
                     reloadCubeQuietly(real.getName());
                 }
             }
+            projectManager.reloadProjectL2Cache(project);
         }
 
         @Override


asfgit:

core-cube/src/main/java/org/apache/kylin/cube/CubeManager.java
```diff
@@ -151,11 +151,13 @@ protected CubeInstance initEntityAfterReload(CubeInstance cube, String resourceN
 
         @Override
         public void onProjectSchemaChange(Broadcaster broadcaster, String project) throws IOException {
-            for (IRealization real : ProjectManager.getInstance(config).listAllRealizations(project)) {
+            ProjectManager projectManager = ProjectManager.getInstance(config);
+            for (IRealization real : projectManager.listAllRealizations(project)) {
```
shaofengshi: Hello, I can't see the difference; can you elaborate on it? Thank you.

core-cube/src/main/java/org/apache/kylin/cube/CubeManager.java
```diff
@@ -151,11 +151,13 @@ protected CubeInstance initEntityAfterReload(CubeInstance cube, String resourceN
...
                 if (real instanceof CubeInstance) {
                     reloadCubeQuietly(real.getName());
                 }
             }
+            projectManager.reloadProjectL2Cache(project);
```
walhl: “onProjectSchemaChange ” method updated all cubeinstance under the given project,  but cubeinstances in ProjectL2Cache  are not updated.  I  added reloading "ProjectL2Cache " after updating cubeinstance.

shaofengshi: Sorry, I missed this line;
In Broadcaster.java line 253, it clears project L2 cache before calling "onProjectSchemaChange", so does it need to reload the project L2 cache again?

walhl: during calling "onProjectSchemaChange", 
there will invoke CubeDescSyncListener.onProjectSchemaChange method. During this method,  "ProjectManager.getInstance(config).listAllRealizations(project) " will lead to loading previous cubeinstance in ProjectL2Cache .  then there will invoke CubeManager.onProjectSchemaChange()  method.

walhl: During invoking "CubeManager.onProjectSchemaChange()" method,  projectL2Cache  did't updated.

shaofengshi: I see, thank you Walhl; If moving "projectManager.reloadProjectL2Cache(project);" to Broadcaster line 256, after all listeners are notified, will that be better?

PR #385, opened 2018-12-11
set JAVA_HOME before server start or stop

shopee-jin:
We believe that the following scenario is fairly common in big data communities:
  1. Have CDH 5.x with JDK 1.7 (/usr/lib/jvm/java-7-oracle-cloudera)
  2. Have JDK 1.8+ for other usage (e.g., kylin 2.5.2+ runs on 1.8+)

The added  set-java-home.sh will try to set JAVA_HOME before start or stop the server.

If JAVA_HOME is not set, under the above scenario, kylin server would try to start with JDK 1.7.  And many jars built after kylin 2.5.0 would NOT run on jdk 1.7.

### DIFF
--- a/build/bin/kylin.sh
+++ b/build/bin/kylin.sh
@@ -30,6 +30,8 @@ fi
 mkdir -p ${KYLIN_HOME}/logs
 mkdir -p ${KYLIN_HOME}/ext
 
+source ${dir}/set-java-home.sh
+
 function retrieveDependency() {
     #retrive $hive_dependency and $hbase_dependency
     source ${dir}/find-hive-dependency.sh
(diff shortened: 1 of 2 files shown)


asfgit:

shaofengshi:
Hi Shopee, thank you for raising this PR. I see the patch, and have a question: it seems this script didn't check the version (1.7 or 1.8), it just export JAVA_HOME (from the path of `java`) if it is absent. So I'm wondering how it solves the problem that CDH ships 1.7 but Kylin expects 1.8?

shaofengshi:
It looks good now. Let me merge it. Thank you Shopee!

PR #387, opened 2018-12-12
APACHE-KYLIN-2932: Simplify the thread model for in-memory cubing

Wayne1c:


asfgit:

PR #390, opened 2018-12-13
KYLIN-3617 Fix for two inconsistent ExecuteManger

hit-lacus:
I am sorry for error JIRA ID, this commit is belong to KYLIN-3617.


hit-lacus:
@nichunen  Hi, could you please review my patch? Thank you.

PR #394, opened 2018-12-15
KYLIN-3720 add column family check when save/update cube desc

wangbo:


asfgit:

server-base/src/main/java/org/apache/kylin/rest/controller/CubeController.java
```diff
@@ -608,6 +613,26 @@ public CubeRequest saveCubeDesc(@RequestBody CubeRequest cubeRequest) {
...
+    private void validateColumnFamily(CubeDesc cubeDesc){
+        Set<String> columnFamilyMetricsSet = Sets.newHashSet();
+        for(HBaseColumnFamilyDesc hBaseColumnFamilyDesc : cubeDesc.getHbaseMapping().getColumnFamily()) {
+            for(HBaseColumnDesc hBaseColumnDesc : hBaseColumnFamilyDesc.getColumns()){
+                for(String columnName : hBaseColumnDesc.getMeasureRefs()){
+                    columnFamilyMetricsSet.add(columnName);
+                }
+            }
+        }
+        for(MeasureDesc measureDesc : cubeDesc.getMeasures()){
```
shaofengshi: Hi bo, the function looks good, but it seems the code is not formatted. Could you reformat it by following the "IDE code formatter" in https://kylin.apache.org/development/dev_env.html ?

Besides, the "merge" commit should be avoid.

Thank you for making Kylin better!

PR #397, opened 2018-12-16
KYLIN-3722 Disable limit push down for complex query

hit-lacus:


asfgit:

hit-lacus:
Local CI test is running.

hit-lacus:
IT test not pass because IT code should be changed.

hit-lacus:
IT test now pass.

query/src/main/java/org/apache/kylin/query/relnode/OLAPLimitRel.java
```diff
@@ -82,7 +82,8 @@ public void implementOLAP(OLAPImplementor implementor) {
         // ignore limit after having clause
         // ignore limit after another limit, e.g. select A, count(*) from (select A,B from fact group by A,B limit 100) limit 10
         // ignore limit after outer aggregate, e.g. select count(1) from (select A,B from fact group by A,B ) limit 10
-        if (!context.afterHavingClauseFilter && !context.afterLimit && !context.afterOuterAggregate) {
+        // ignore limit after join
+        if (!context.afterHavingClauseFilter && !context.afterLimit && !context.afterOuterAggregate && !context.afterJoin) {
```
shaofengshi: My concern is, this change may make many normal queries (join with the limit) much inefficient.

hit-lacus: Good point, I am looking for better solution now.

hit-lacus:
1. add a switch `kylin.storage.limit-push-down-enable` which could turn off limit push down
2.  turn off limit push down when multi olap context related
3. add a new sql in IT

Local CI test passed.

PR #399, opened 2018-12-17
KYLIN-3720 add column family check when save/update cube desc

wangbo:
check code-style


asfgit:

PR #402, opened 2018-12-18
 KYLIN-3724 Kylin IT test sql is unreasonable

hit-lacus:
local ci test pass


asfgit:

PR #405, opened 2018-12-18
KYLIN-1819 Fix exception swallowed when start DefaultScheduler fail

kyotoYaho:


asfgit:

PR #407, opened 2018-12-19
KYLIN-3540 estimate the row counts of source cuboids which are not built & remove mandatory cuboids recommendation

kyotoYaho:


asfgit:

PR #408, opened 2018-12-19
KYLIN-3655 reinitialize CubeInstance when clear segments

kyotoYaho:


asfgit:

core-cube/src/main/java/org/apache/kylin/cube/CubeManager.java
```diff
@@ -256,6 +256,23 @@ public CubeInstance createCube(CubeInstance cube, String projectName, String own
...
+     */
+    public CubeInstance clearSegments(CubeInstance cube) throws IOException {
+        try (AutoLock lock = cubeMapLock.lockForWrite()) {
+            cube = cube.latestCopyForWrite(); // get a latest copy
+            CubeUpdate update = new CubeUpdate(cube);
+            update.setToRemoveSegs(cube.getSegments().toArray(new CubeSegment[cube.getSegments().size()]));
+            update.setCuboids(Maps.<Long, Long> newHashMap());
+            update.setCuboidsRecommend(Sets.<Long> newHashSet());
+            update.setUpdateTableSnapshotPath(Maps.<String, String> newHashMap());
+            update.setCreateTimeUTC(System.currentTimeMillis());
```
shaofengshi: What's the reason to update the "createTimeUTC"?

core-cube/src/main/java/org/apache/kylin/cube/CubeManager.java
```diff
@@ -394,6 +411,14 @@ private void setCubeMember(CubeInstance cube, CubeUpdate update) {
                 cube.putSnapshotResPath(lookupSnapshotPathEntry.getKey(), lookupSnapshotPathEntry.getValue());
             }
         }
+
+        if (update.getCreateTimeUTC() >= 0) {
+            cube.setCreateTimeUTC(update.getCreateTimeUTC());
```
shaofengshi: Same as above, why to reset the "createTimeUTC" on an update, that is a little strange.

core-cube/src/main/java/org/apache/kylin/cube/CubeManager.java
kyotoYaho: Now the mapping of CubeDesc & CubeInstance is one to one. It may not be necessary to have create time UTC for CubeInstance. Here, the create time UTC is regarded as the reinitialization time.

PR #409, opened 2018-12-19
KYLIN-2972 CacheKey from SQLRequest should ignore the case of project…

kyotoYaho:
… name


asfgit:

server-base/src/main/java/org/apache/kylin/rest/request/SQLRequest.java
```diff
@@ -105,7 +105,7 @@ public Object getCacheKey() {
             return cacheKey;
 
         cacheKey = Lists.newArrayList(sql.replaceAll("[ ]", " ") //
-                , project //
+                , getNormProject() //
```
shaofengshi: Can we normalize the value to upper case when the request is received? Then downstream doesn't need to handle the case each time.

PR #412, opened 2018-12-23
KYLIN-3730 TableMetadataManager.reloadSourceTableQuietly is wrong

hit-lacus:
https://issues.apache.org/jira/projects/KYLIN/issues/KYLIN-3730?filter=myopenissues


asfgit:

PR #419, opened 2018-12-27
fix query no data when partion column as filter 

javalife0312:
kylin on druid,when query like [select * from tb where dt='yyyyMMdd'] will no data return
fix query no data when partion column as filter


asfgit:

javalife0312:
kylin on druid,when query like [select * from tb where dt='yyyyMMdd'] will no data return
fix query no data when partion column as filter

kangkaisen:
LGTM. Thanks you!

PR #428, opened 2019-01-11
KYLIN-3767 Print the malformed JSON data consumed from Kafka Topic

TempleZhou:


asfgit:

PR #429, opened 2019-01-15
KYLIN-3770 ZipFileUtils is not compatible CubeMetaIngester

TempleZhou:


asfgit:

PR #435, opened 2019-01-18
when dimvalue`s len > short.max_value then [Extract Fact Table Distinct Columns] err

javalife0312:
https://issues.apache.org/jira/browse/KYLIN-3775

### DIFF
--- a/core-dictionary/src/main/java/org/apache/kylin/dict/TrieDictionary.java
+++ b/core-dictionary/src/main/java/org/apache/kylin/dict/TrieDictionary.java
@@ -94,7 +94,7 @@ public class TrieDictionary<T> extends CacheDictionary<T> {
             this.maxValueLength = headIn.readInt();
             if (maxValueLength < 0) {
                 throw new IllegalStateException("maxValueLength is negative (" + maxValueLength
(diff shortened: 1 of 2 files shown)


asfgit:

core-dictionary/src/main/java/org/apache/kylin/dict/TrieDictionary.java
```diff
@@ -89,6 +89,35 @@ public TrieDictionary(byte[] trieBytes) {
...
+        try {
+            this.maxValueLength = headIn.readInt();
+            if (maxValueLength < 0) {
+                throw new IllegalStateException("maxValueLength is negative (" + maxValueLength
+                        + "). Dict value is too long, whose length is larger than " + Integer.MAX_VALUE);
+            }
+            String converterName = headIn.readUTF();
+
+            return converterName;
+        } catch (Exception e) {
```
nichunen: When will this happen?

nichunen:
@javalife0312 Hi, can you open a jira, and describe the issue troubled  you in detail?

nichunen:
After discussed with @javalife0312  
We think this is not a common case, and there is workaround way to deal with long dimensions.
Close this pr

PR #436, opened 2019-01-19
KYLIN-3716 FastThreadLocal replaces ThreadLocal

SteNicholas:
In kylin query engine,QuerySevice acquires OLAPContext through ThreadLocal.In certain research,the development that FastThreadLocal replaces ThreadLocal to store thread OLAPContext is substantial performance improvement.


asfgit:

SteNicholas:
@shaofengshi I did maven test and rename the classes of threadlocal package that causes some exception you referred.I have already return to the origin class name and checked maven test passed.

SteNicholas:
@shaofengshi Yeah,it is possible to add the dependency on netty and then use netty's "FastThreadLocal" class in Kylin.I simplify FastThreadLocal by using InternalThreadLocal and avoid more dependency in Kylin.Therefore, there is two ways to improve ThreadLocal performance by different aspects.

shaofengshi:
Hi Nicholas, I merge the change by merging the two commits into one, and modified the commit message to follow Apache rule. Thanks for your patience!

PR #437, opened 2019-01-20
KYLIN-3776 Float type in MySQL not properly converted to HIVE Double …

hit-lacus:
In Hive DDL manual, the precision/scale of double and float is not configurable


hit-lacus:
Local CI test pass.

shaofengshi:
LGTM

shaofengshi:
How about "return typeName.equals("decimal") || typeName.equals("numeric"); "?

PR #441, opened 2019-01-22
Kylin 3494: Build cube with spark reports ArrayIndexOutOfBoundsException

lshmouse:
This issue is introduced in the commit:
```
commit e0f29e29164d44581c133314ac01c419e4277c35
Author: shaofengshi <shaofengshi@apache.org>
Date:   Thu May 31 11:17:49 2018 +0800

    KYLIN-3137 Spark cubing refine
```
After the commit, kylin will use sc.sequenceFile to read sequence file and split the field using string splitting. It will lose null column fields and cause ArrayIndexOutOfBoundsException in BaseCuboidBuilder.
So just rolling the back the changes in commit: e0f29e29


asfgit:

asfgit:

engine-spark/src/main/java/org/apache/kylin/engine/spark/SparkUtil.java
```diff
@@ -143,30 +139,7 @@ public static void modifySparkHadoopConfiguration(SparkContext sc) throws Except
...
-                    @Override
-                    public String[] call(Text text) throws Exception {
-                        String s = Bytes.toString(text.getBytes(), 0, text.getLength());
-                        return s.split(BatchConstants.SEQUENCE_FILE_DEFAULT_DELIMITER);
-                    }
-                });
-    }
-
-    private static JavaRDD<String[]> getOtherFormatHiveInput(JavaSparkContext sc, String hiveTable) {
+    public static JavaRDD<String[]> getHiveInput(JavaSparkContext sc, String hiveTable) {
```
shaofengshi: You deleted this logic, so all case will go to hive context to parse, that will add a dependency on Hive, which is we want to avoid...

PR #442, opened 2019-01-22
KYLIN-3780: Add built instance in Job info

lshmouse:
In DistributedScheduler, it's hard to known which machine the kylin job is running on.
But this info is helpful to debug the failed jobs.
So we add the  job built instance info in job info and kylin web ui.


lshmouse:
@shaofengshi not find this PR in the master branch. Is it missing?

lshmouse:
@shaofengshi  Sorry. Found it. Thanks~
It is not easy to found because my commit message of pr is not formatted.

PR #443, opened 2019-01-23
KYLIN-3759 & KYLIN-3744

hit-lacus:
The purpose of this PR is as following:
1. Fix a small bug in mapreduce job when lambda is enabled.
2. Add enhancement for reassign of Coordinator
3. Add unit test for for reassign of Coordinator
4. Add some javadoc

### DIFF
--- a/stream-coordinator/src/main/java/org/apache/kylin/stream/coordinator/Coordinator.java
+++ b/stream-coordinator/src/main/java/org/apache/kylin/stream/coordinator/Coordinator.java
@@ -20,10 +20,13 @@ package org.apache.kylin.stream.coordinator;
 
 import java.io.Closeable;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentMap;
@@ -32,7 +35,10 @@ import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
 import com.google.common.collect.Collections2;
 import org.apache.curator.framework.CuratorFramework;
@@ -42,6 +48,7 @@ import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.kylin.common.KylinConfig;
 import org.apache.kylin.common.util.HadoopUtil;
+import org.apache.kylin.common.util.JsonUtil;
 import org.apache.kylin.common.util.Pair;
 import org.apache.kylin.common.util.ServerMode;
 import org.apache.kylin.cube.CubeInstance;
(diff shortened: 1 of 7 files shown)


asfgit:

asfgit:

stream-coordinator/src/main/java/org/apache/kylin/stream/coordinator/Coordinator.java
```diff
@@ -132,14 +143,37 @@ private Coordinator() {
...
+        this.jobStatusChecker = new StreamingBuildJobStatusChecker();
+        this.streamingJobCheckExecutor = Executors.newScheduledThreadPool(1,
+                new NamedThreadFactory("streaming_job_status_checker"));
+        if (ServerMode.SERVER_MODE.canServeStreamingCoordinator()) {
+            start();
+        }
+        this.isLead = true;
+    }
+
+    public static synchronized Coordinator getInstance() {
```
allenma: late initialization here for unit test?
no need synchronized on method level

hit-lacus: OK, I understand.

stream-coordinator/src/main/java/org/apache/kylin/stream/coordinator/Coordinator.java
```diff
@@ -407,26 +448,43 @@ public void doReassign(CubeInstance cubeInstance, CubeAssignment preAssignments,
...
                 } catch (IOException e1) {
-                    logger.error(
-                            "fail to start consumers for cube:" + cubeName + " replicaSet:" + rs.getReplicaSetID(), e1);
+                    logger.error("fail to start consumers for cube:" + cubeName + " replicaSet:" + rs.getReplicaSetID(),
+                            e1);
                 }
             }
-            throw new RuntimeException(e);
+            if (needRollback.isEmpty()) {
+                throw new ClusterStateInconsistentException(ClusterState.ROLLBACK_SUCCESS,
```
allenma: when needRollback is empty, the rebalance is failure, but the cluster is still consistent, so should not throw ClusterStateInconsistentException

stream-coordinator/src/main/java/org/apache/kylin/stream/coordinator/Coordinator.java
```diff
@@ -478,24 +544,61 @@ public void doReassign(CubeInstance cubeInstance, CubeAssignment preAssignments,
...
+                throw new ClusterStateInconsistentException(ClusterState.ROLLBACK_FAILED, TransactionStep.START_NEW,
+                        failedInfo, e);
+            } else if (!rollbackAssigned.isEmpty()) {
+                throw new ClusterStateInconsistentException(ClusterState.ROLLBACK_FAILED, TransactionStep.ASSIGN_NEW,
+                        failedInfo, e);
+            } else if (!failedReceiver.isEmpty()) {
+                throw new ClusterStateInconsistentException(ClusterState.ROLLBACK_FAILED,
+                        TransactionStep.MAKE_IMMUTABLE, failedInfo, e);
+            } else {
+                throw new ClusterStateInconsistentException(ClusterState.ROLLBACK_SUCCESS, TransactionStep.ASSIGN_NEW,
```
allenma: Rollback success, the cluster state should be consistent. And the state should be cube level
And should we put cube's inconsistent state somewhere, and return to client so that we can show the state on GUI.

stream-coordinator/src/main/java/org/apache/kylin/stream/coordinator/Coordinator.java
hit-lacus: Maybe my code is not clear enough, I use `ClusterState.ROLLBACK_SUCCESS` to indicate that the cluster state is consistent but reassign is failed. I will modify this place later.

stream-coordinator/src/main/java/org/apache/kylin/stream/coordinator/Coordinator.java
hit-lacus: > Rollback success, the cluster state should be consistent.

As I explained above,  I use ClusterState.ROLLBACK_SUCCESS to indicate that the cluster state is consistent but reassign is failed.

> And should we put cube's inconsistent state somewhere, and return to client so that we can show the state on GUI.

Yes, I agree. If this patch could be approved, I will try to store cube inconsistent  state in zk, and maybe try to send email or display warning somewhere.

hit-lacus: The name of `ClusterStateInconsistentException` is not clear, maybe change to `ClusterStateException` is better. And it should add a field `cubeName` to that class.

PR #446, opened 2019-01-25
KYLIN 3790 return wrong error code when command executor is interrupted

chenzhx:


asfgit:

asfgit:

PR #447, opened 2019-01-25
KYLIN-3494: Build cube with spark reports ArrayIndexOutOfBoundsException

lshmouse:
The default string split method will remove empty values.  This will lose null column fields and cause ArrayIndexOutOfBoundsException in BaseCuboidBuilder.

Just fix it using string split with -1 limit.
See: https://stackoverflow.com/questions/14602062/java-string-split-removed-empty-values


asfgit:

asfgit:

lshmouse:
@shaofengshi  Please help to review the patch, thanks~

PR #466, opened 2019-02-12
KYLIN-3786 Add integration test for realtime-streaming

Wayne1c:


asfgit:

Wayne1c:
Test data from https://github.com/hit-lacus/hit-lacus.github.io/issues/13#issuecomment-450480992
1, Fixed 10000 records, on the one hand, load them to h2 database for the comparison of query result; on the other hand, send them to mocked Kafka server for receiver to consume.
2, StreamingServer and Coordinator communicate by calling the method of each other instead of throughing http request.

PR #470, opened 2019-02-13
#KYLIN-3814, add pause interval before the job retry

ZhengshuaiPENG:


asfgit:

PR #472, opened 2019-02-20
KYLIN-3819 kylin.sh run in the foreground

TempleZhou:


asfgit:

PR #478, opened 2019-02-22
KYLIN-3824

Sidonet:
For environment with low RAM, and huge Cube, remove List to avoid java.lang.OutOfMemory and iterate by Iterator.


asfgit:

Sidonet:
image

PR #479, opened 2019-02-24
KYLIN-3826 MergeCuboidJob only uploads necessary segment's dictionary

shaofengshi:


asfgit:

PR #482, opened 2019-02-25
KYLIN-3828 Fix heading empty string be ignored in StringUtil.join

nichunen:


asfgit:

PR #483, opened 2019-02-25
KYLIN-3816 Make CI cover streaming table join case

Wayne1c:


asfgit:

PR #485, opened 2019-02-26
KYLIN-3829 Chinese document on the kylin official site about sample c…

GinaZhai:
…ube got the sql wrong


asfgit:

PR #490, opened 2019-02-27
Added small Unit Tests

TheRealHaui:
Added small Unit Tests


asfgit:

shaofengshi:
Hi Michael, thanks for contributing to Kylin! The change seems not passing the unit test, could you please take a look? Just let me know if you need support.

TheRealHaui:
@shaofengshi,
thanks for your kind reply!
Oversaw that your build servers complained.
The problem was a checkstyle rule.
Removed .* imports from tests.

PR #493, opened 2019-02-28
KYLIN-3836 fix Kylin StringUtil.join() may cause NPE if iterator is e…

Wayne1c:
…mpty


asfgit:

PR #496, opened 2019-03-04
KYLIN-3838 Fix retry mechanism is invalid when build with spark

Wayne1c:


asfgit:

PR #507, opened 2019-03-10
KYLIN-3857 Disable quote when needed

hit-lacus:
1. add quote for create table ddl
2. add switch for turn off quote

### DIFF
--- a/core-common/src/main/java/org/apache/kylin/common/KylinConfigBase.java
+++ b/core-common/src/main/java/org/apache/kylin/common/KylinConfigBase.java
@@ -873,6 +873,14 @@ public abstract class KylinConfigBase implements Serializable {
         return r;
     }
 
+    public boolean enableHiveDdlQuote(){
+        return Boolean.parseBoolean(getOptional("kylin.source.hive.quote-enabled", TRUE));
+    }
+
+    public String getQuoteCharacter(){
+        return getOptional("kylin.source.quote.character", "`");
+    }
+
     /**
      * was for route to hive, not used any more
      */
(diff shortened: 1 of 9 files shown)


asfgit:

hit-lacus:
Local CI Test passed.

core-common/src/main/java/org/apache/kylin/common/KylinConfigBase.java
```diff
@@ -982,6 +982,14 @@ public int getDefaultDecimalScale() {
         }
     }
 
+    public boolean enableHiveDdlQuote(){
+        return Boolean.parseBoolean(getOptional("kylin.source.hive.enable.quote", TRUE));
```
shaofengshi: Please follow the config naming convension, "kylin.source.hive.quote-enabled" is a better name, like "kylin.engine.spark.sanity-check-enabled"

hit-lacus: I have fix this.

core-common/src/main/java/org/apache/kylin/common/KylinConfigBase.java
```diff
@@ -982,6 +982,14 @@ public int getDefaultDecimalScale() {
         }
     }
 
+    public boolean enableHiveDdlQuote(){
+        return Boolean.parseBoolean(getOptional("kylin.source.hive.quote-enabled", TRUE));
+    }
+
+    public static String getQuoteCharacter(){
```
shaofengshi: Why not make this configurable?

hit-lacus: The quote character used by Hive is **`**, I think it wouldn't be changed easily.

hit-lacus: Make it configurable do not make harm, I will change it.

PR #509, opened 2019-03-11
KYLIN-3864 Provide a function to judge whether the os type is Mac os x or not

yanghua:
JIRA ref : https://issues.apache.org/jira/browse/KYLIN-3864

### DIFF
--- a/build/bin/util.sh
+++ b/build/bin/util.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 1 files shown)


asfgit:

shaofengshi:
Hi Hua, which script will use this?

yanghua:
Hi @shaofengshi  Now, In kylin project, there are many places where the code would check the os type is **Darwin** (Mac os X) such as `download-spark.sh`, `download-flink.sh`, `tool-wrapper.sh`. And there are some places where Kylin is not compatible with Mac os X (it does not check the os type), it caused some errors, such as [KYLIN-3756](https://issues.apache.org/jira/browse/KYLIN-3756) and [KYLIN-3757](https://issues.apache.org/jira/browse/KYLIN-3757). This util method provides a general function to verify whether the current os type is Mac os X or not.

By providing this function, we can do generic checking for the different `like-UNIX` os, and fix those compatible issues existing in the shell script.

shaofengshi:
I see, there will be subsequent changes based on this. I will merge it.

PR #510, opened 2019-03-12
KYLIN-3867 Enable JDBC to use key store & trust store for https connection

kyotoYaho:


asfgit:

PR #511, opened 2019-03-12
KYLIN-3830 return wrong result when 'SELECT SUM(dim1)' without set a …

Qsimple:
just remove FUNC_SUM in BUILT_IN_AGGREGATIONS, which make the kylin doesn't support query sum(dim1) without set a relative metric of dim1

### DIFF
--- a/core-metadata/src/main/java/org/apache/kylin/metadata/model/FunctionDesc.java
+++ b/core-metadata/src/main/java/org/apache/kylin/metadata/model/FunctionDesc.java
@@ -68,7 +68,6 @@ public class FunctionDesc implements Serializable {
         BUILT_IN_AGGREGATIONS.add(FUNC_COUNT);
         BUILT_IN_AGGREGATIONS.add(FUNC_MAX);
         BUILT_IN_AGGREGATIONS.add(FUNC_MIN);
-        BUILT_IN_AGGREGATIONS.add(FUNC_SUM);
         BUILT_IN_AGGREGATIONS.add(FUNC_COUNT_DISTINCT);
         BUILT_IN_AGGREGATIONS.add(FUNC_PERCENTILE);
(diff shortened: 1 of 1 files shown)


asfgit:

shaofengshi:
Thanks for the pr; With this change, Kylin will reject such query, is that true?

Qsimple:
Yes, I have tested it. when query by `SELECT SUM(SELLER_ID) FROM KYLIN_SALES;`, kylin throw
```
No realization found for OLAPContext, CUBE_NOT_READY, CUBE_NOT_READY, CUBE_UNMATCHED_AGGREGATION[FunctionDesc [expression=SUM, parameter=1_317b3789:DEFAULT.KYLIN_SALES.SELLER_ID, returnType=null]], CUBE_NOT_READY, CUBE_NOT_READY, CUBE_NOT_READY, CUBE_NOT_READY, CUBE_NOT_READY, CUBE_NOT_READY, CUBE_NOT_READY, rel#59:OLAPTableScan.OLAP.[](table=[DEFAULT, KYLIN_SALES],ctx=,fields=[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]) while executing SQL: "SELECT SUM(SELLER_ID) FROM KYLIN_SALES LIMIT 50000"
```   
Because I have found `FunctionDesc.BUILT_IN_AGGREGATIONS` only used in CubeCapabilityChecker:tryDimensionAsMeasure:235, which judge the FunctionDesc(SUM,COUNT,MAX,MIN,PERCENTILE,COUNT_DISTINCT) can be used as Dimension as measure.  
As we discussed before, the SUM can't be calculated right. So I remove it to reject such query. Although the error message seems obscure, the commit can fix this bug with least change.

PR #513, opened 2019-03-13
KYLIN-3860 Add doc about configuration of kylin.web.hide-measures

Qsimple:


asfgit:

PR #514, opened 2019-03-13
KYLIN-3868 Implement Flink batch cubing job builder

yanghua:
JIRA ref : https://issues.apache.org/jira/browse/KYLIN-3868


asfgit:

PR #516, opened 2019-03-14
KYLIN-3874 "Convert Cuboid Data to HFile" failed when HBase and MR use different HDFS clusters

guangxuCheng:
Detail: https://issues.apache.org/jira/browse/KYLIN-3874


asfgit:

shaofengshi:
Guangxu, thank you for the PR! We will review it soon.

guangxuCheng:
Sorry, There is a small problem here. After modification, it may cause the configuration in "kylin_job_conf.xml" to be overwritten. Let me prepare a new patch, thanks!

guangxuCheng:
Re-review the logic of this block, we only need to add the information about the nameservice of HBase's HDFS to the configuration. The patch has been updated.Thanks

PR #519, opened 2019-03-15
KYLIN-3262 add a copy phase for copying jacoco related jars to dev-support folder

kyotoYaho:


asfgit:

PR #520, opened 2019-03-15
KYLIN-3850 Flink cubing step : merge by layer

yanghua:
JIRA ref : https://issues.apache.org/jira/browse/KYLIN-3850


asfgit:

PR #521, opened 2019-03-15
KYLIN-3876 Fix flink config option wrong key mapping

yanghua:
JIRA ref : https://issues.apache.org/jira/browse/KYLIN-3876


asfgit:

PR #530, opened 2019-03-18
KYLIN-3886:  Missing argument for options for yarn command

lshmouse:
See: https://issues.apache.org/jira/browse/KYLIN-3886

The app id may be empty which cause the yarn command failed for missing argument.
Checking the app id before executing yarn command


asfgit:

asfgit:

PR #533, opened 2019-03-19
kylin-3812 optimize the child CompareTupleFilter in a CompareTupleFilter

kyotoYaho:


asfgit:

asfgit:

core-metadata/src/main/java/org/apache/kylin/metadata/filter/CompareTupleFilter.java
```diff
@@ -283,6 +289,20 @@ public TupleFilter acceptOptimizeTransformer(FilterOptimizeTransformer transform
...
+    private TupleFilter optimizeChildCompareTupleFilter(CompareTupleFilter child) {
+        FilterOptimizeTransformer transformer = new FilterOptimizeTransformer();
+        TupleFilter result = child.acceptOptimizeTransformer(transformer);
+        if (result == ConstantTupleFilter.TRUE) {
+            // use string instead of boolean since it's encoded as string
+            result = new ConstantTupleFilter("true");
+        } else if (result == ConstantTupleFilter.FALSE) {
+            result = new ConstantTupleFilter("false");
+        } else {
+            this.isNormal = false;
```
vaful: this means that TupleFilters excepted AlwaysTrue or AlwaysFalse will not pushdown?

core-metadata/src/main/java/org/apache/kylin/metadata/filter/CompareTupleFilter.java
```diff
@@ -74,6 +77,9 @@ private CompareTupleFilter(CompareTupleFilter another) {
 
     @Override
     public void addChild(TupleFilter child) {
+        if (child instanceof CompareTupleFilter) {
```
vaful: plz add some test case in UT or IT

core-metadata/src/main/java/org/apache/kylin/metadata/filter/CompareTupleFilter.java
kyotoYaho: Why do you think AlwaysTrue or AlwaysFalse cannot be pushed down? Current CompareTupleFilter does not well support the case that both children are CompareTupleFilter. Therefore, push down will be disabled. This behavior is not changed.

core-metadata/src/main/java/org/apache/kylin/metadata/filter/CompareTupleFilter.java
kyotoYaho: We need to refine our IT and add more test cases. To test this case, it's necessary to add a column with boolean type.

PR #535, opened 2019-03-20
KYLIN-3895 Failed to register new MBean when "kylin.server.query-metrics-enabled" set true

guangxuCheng:
see: https://issues.apache.org/jira/browse/KYLIN-3895

### DIFF
--- a/server-base/src/main/java/org/apache/kylin/rest/metrics/QueryMetricsFacade.java
+++ b/server-base/src/main/java/org/apache/kylin/rest/metrics/QueryMetricsFacade.java
@@ -79,12 +79,11 @@ public class QueryMetricsFacade {
             return;
 
         String projectName = sqlRequest.getProject();
-        String cubeName = sqlResponse.getCube();
-
         update(getQueryMetrics("Server_Total"), sqlResponse);
-
         update(getQueryMetrics(projectName), sqlResponse);
 
+        String cube = sqlResponse.getCube();
+        String cubeName = cube.replace("=", "->");
         String cubeMetricName = projectName + ",sub=" + cubeName;
         update(getQueryMetrics(cubeMetricName), sqlResponse);
     }


asfgit:

asfgit:

server-base/src/main/java/org/apache/kylin/rest/metrics/QueryMetricsFacade.java
```diff
@@ -79,12 +79,11 @@ private static void updateMetricsToLocal(SQLRequest sqlRequest, SQLResponse sqlR
...
 
         String projectName = sqlRequest.getProject();
-        String cubeName = sqlResponse.getCube();
-
         update(getQueryMetrics("Server_Total"), sqlResponse);
-
         update(getQueryMetrics(projectName), sqlResponse);
 
+        String cube = sqlResponse.getCube();
+        String cubeName = cube.replace("=", "->");
```
shaofengshi: Do you mean, when the Cube name includes "=", it will be rejected? Thx!

guangxuCheng: The format of sqlResponse.getCube() is "CUBE[name=KYLIN_HIVE_METRICS_JOB_QA]", the subname of cubeMetricName always contains "=" which is not conform to the naming convention of mbean. Metrics2 has done a similar operation. See: https://github.com/apache/kylin/blob/d9b028b24a0ad9ca09d4d4f1f6ee4de1e59ccd6a/server-base/src/main/java/org/apache/kylin/rest/metrics/QueryMetrics2Facade.java#L64

shaofengshi: I see, thank you for the information!

PR #538, opened 2019-03-20
KYLIN-3898 Cube level properties are ineffective in the some build steps

TempleZhou:


asfgit:

asfgit:

shaofengshi:
Hi Temple, I think the more clear way is in :

org.apache.kylin.job.JoinedFlatTable#generateCreateTableStatement(org.apache.kylin.metadata.model.IJoinedFlatTableDesc, java.lang.String)

Change ` String storageFormat = flatDesc.getDataModel().getConfig().getFlatTableStorageFormat();`

to 
` String storageFormat = flatDesc.getSegment().getConfig().getFlatTableStorageFormat();`

Which is to say, here using segment level config (which is also the cube level) is better than the model level config. Just let me know your idea.

TempleZhou:
> Hi Temple, I think the more clear way is in :
> 
> org.apache.kylin.job.JoinedFlatTable#generateCreateTableStatement(org.apache.kylin.metadata.model.IJoinedFlatTableDesc, java.lang.String)
> 
> Change ` String storageFormat = flatDesc.getDataModel().getConfig().getFlatTableStorageFormat();`
> 
> to
> ` String storageFormat = flatDesc.getSegment().getConfig().getFlatTableStorageFormat();`
> 
> Which is to say, here using segment level config (which is also the cube level) is better than the model level config. Just let me know your idea.

Oh ~ you are right, changing the dataModel is a foolish way as there are different cube use same model.
I have tested the change you advised, no problem.

PR #543, opened 2019-03-21
 KYLIN-3885: Build dimension dictionary job costs too long when using Spark fact distinct

lshmouse:
See: https://issues.apache.org/jira/browse/KYLIN-3893

The issue is that the ColumnDFSFile generated by Spark fact distinct is not sorted, which cause too many write/delete/read slice files when building global dictionary.

### DIFF
--- a/core-dictionary/src/main/java/org/apache/kylin/dict/global/AppendTrieDictionaryBuilder.java
+++ b/core-dictionary/src/main/java/org/apache/kylin/dict/global/AppendTrieDictionaryBuilder.java
@@ -22,6 +22,8 @@ import org.apache.kylin.common.util.BytesUtil;
 import org.apache.kylin.dict.AppendTrieDictionary;
 import org.apache.kylin.dict.BytesConverter;
 import org.apache.kylin.dict.StringBytesConverter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


asfgit:

asfgit:

core-dictionary/src/main/java/org/apache/kylin/dict/global/AppendTrieDictionaryBuilder.java
```diff
@@ -86,7 +89,13 @@ public void addValue(String value) throws IOException {
         }
         checkState(sliceFileMap.firstKey().equals(AppendDictSliceKey.START_KEY), "first key should be \"\", but got \"%s\"", sliceFileMap.firstKey());
 
-        AppendDictSliceKey nextKey = sliceFileMap.floorKey(AppendDictSliceKey.wrap(valueBytes));
+        AppendDictSliceKey key = AppendDictSliceKey.wrap(valueBytes);
+        if (preKey != null && key.compareTo(preKey) < 0) {
```
shaofengshi: I'm afraid this check for each key will downgrade the performance of global dict building.

lshmouse: This code is just used for verifying the issue. I can remove it later or add a flag for it.

shaofengshi:
Merged together with PR 565; thank you Shaohui!

PR #545, opened 2019-03-22
KYLIN-3902 fix JoinDesc in case of same fact column with multiple loo…

kyotoYaho:
…kup columns


asfgit:

asfgit:

PR #549, opened 2019-03-25
KYLIN-3905 Enable shrunken dictionary default

hit-lacus:
In dev mail list's discussion, I suggest to enable shrunken dictionary by default, and received some dev's aggrement. 
When using bitmap measure on a large cardinality column(require global dictionaty), build base cuboid step need frequent cache swap so it cannot finished within a reasonable period.
When shrunken dictionary enabled, a new step will be added to build separated dictionary for each `InputSplit`, Mapper of *BuildBaseCuboid* step only has to fetch a smaller dictionary for itself, instead of a larger global dictionary. It will reduce cache swap and make *BuildBaseCuboid* step run as quicker as possible.


asfgit:

asfgit:

PR #560, opened 2019-03-27
KYLIN-3918: Add project name in cube and job pages

lshmouse:
See: https://issues.apache.org/jira/browse/KYLIN-3918


asfgit:

asfgit:

nichunen:
@lshmouse Hi, why should project name should be shown on each job even if the job name can be seen on the project drop down list?

PR #561, opened 2019-03-27
KYLIN-3915 Replace dead HADOOP_CLASSPATH in FlinkExecutable

yanghua:
JIRA ref : https://issues.apache.org/jira/browse/KYLIN-3915

### DIFF
--- a/engine-flink/src/main/java/org/apache/kylin/engine/flink/FlinkExecutable.java
+++ b/engine-flink/src/main/java/org/apache/kylin/engine/flink/FlinkExecutable.java
@@ -45,6 +45,7 @@ import org.apache.kylin.metadata.model.Segments;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -151,8 +152,7 @@ public class FlinkExecutable extends AbstractExecutable {
 
             String jars = this.getParam(JARS);
 
-            String hadoopConf = null;
-            hadoopConf = System.getProperty("kylin.hadoop.conf.dir");
+            String hadoopConf = System.getProperty("kylin.hadoop.conf.dir");
 
             if (StringUtils.isEmpty(hadoopConf)) {
                 throw new RuntimeException(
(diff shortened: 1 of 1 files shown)


asfgit:

asfgit:

engine-flink/src/main/java/org/apache/kylin/engine/flink/FlinkExecutable.java
```diff
@@ -171,35 +171,37 @@ protected ExecuteResult doWork(ExecutableContext context) throws ExecuteExceptio
             Segments<CubeSegment> mergingSeg = cube.getMergingSegments(segment);
             dumpMetadata(segment, mergingSeg);
 
+            Map<String, String> flinkConfs = config.getFlinkConfigOverride();
+            String hadoopClasspathEnv = flinkConfs.get(ENV_HADOOP_CLASSPATH_KEY);
+
             StringBuilder sb = new StringBuilder();
             if (Shell.osType == Shell.OSType.OS_TYPE_WIN) {
-                sb.append("set HADOOP_CONF_DIR=%s && export HADOOP_CLASSPATH=/usr/hdp/2.4.0.0-169/hadoop/ && %s/bin/flink run -m yarn-cluster ");
+                sb.append("set HADOOP_CONF_DIR=%s && export HADOOP_CLASSPATH=%s && %s/bin/flink run -m yarn-cluster ");
```
shaofengshi: should "HADOOP_CLASSPATH" also use "set" instead of "export"?

core-common/src/main/resources/kylin-defaults.properties
```diff
@@ -336,6 +336,7 @@ kylin.engine.spark-conf-mergedict.spark.memory.fraction=0.2
 kylin.engine.flink-conf.jobmanager.heap.size=2G
 kylin.engine.flink-conf.taskmanager.heap.size=4G
 kylin.engine.flink-conf.taskmanager.numberOfTaskSlots=1
+kylin.engine.flink-conf.env.hadoopclasspath=/usr/hdp/2.4.0.0-169/hadoop/
```
shaofengshi: Can this Hadoop path be detected at runtime? It is difficult for the user to configure this. Please refer to the "find-hadoop-conf-dir.sh"

engine-flink/src/main/java/org/apache/kylin/engine/flink/FlinkExecutable.java
yanghua: Yes, Good catch!

core-common/src/main/resources/kylin-defaults.properties
yanghua: OK, I will rethink this.

PR #571, opened 2019-03-29
KYLIN-3922 Fix coprocessor can not be updated

nichunen:


asfgit:

PR #576, opened 2019-04-01
KYLIN-3928 pushdown only return one column

javalife0312:
pushdown only return one column


asfgit:

asfgit:

javalife0312:
my test error

PR #583, opened 2019-04-06
KYLIN-3788  modify the time conversion time zone of the kafka streaming access; It solves the problems that the kafka stream time is different from the real time

zhaojintaozhao:
This submission modifies the time conversion time zone of the kafka streaming access; It solves the problems that the kafka stream time is different from the real time.
For example, if the time zone if "GMT+8",  the hour_time of kafka flow task is "2019-04-05 22:00:00"; in fact the real time is "2019-04-06 06:00:00"

### DIFF
--- a/core-common/src/main/java/org/apache/kylin/common/util/DateFormat.java
+++ b/core-common/src/main/java/org/apache/kylin/common/util/DateFormat.java
@@ -18,7 +18,6 @@
 package org.apache.kylin.common.util;
 
 import java.text.ParseException;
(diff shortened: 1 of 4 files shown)


asfgit:

asfgit:

zhaojintaozhao:
In my Kylin system, this pull requests solve the  problem of time errors in kafka streaming tasks. I hope my code can  be fitted into the kylin trunk brach.

PR #585, opened 2019-04-08
KYLIN-3936 MR/Spark task will still run after the job is stopped

guangxuCheng:
The command "pause" only sets status of the job to "stopped" and does not reset the status of the subtask.https://github.com/apache/kylin/blob/e9dfaf9a5465ff3f6f3bfa3460ca8b2adb8c6617/server-base/src/main/java/org/apache/kylin/rest/service/JobService.java#L662-L666
So, In SparkExecutable, we can't get the real status of the running task. https://github.com/apache/kylin/blob/e9dfaf9a5465ff3f6f3bfa3460ca8b2adb8c6617/engine-spark/src/main/java/org/apache/kylin/engine/spark/SparkExecutable.java#L293-L300

### DIFF
--- a/core-job/src/main/java/org/apache/kylin/job/execution/ExecutableManager.java
+++ b/core-job/src/main/java/org/apache/kylin/job/execution/ExecutableManager.java
@@ -375,7 +375,8 @@ public class ExecutableManager {
             } else {
                 logger.warn("The job " + jobId + " has been discarded.");
             }
-            return;
+            throw new IllegalStateException(
+                "The job " + job.getId() + " has already been finished and cannot be discarded.");
         }
         if (job instanceof DefaultChainedExecutable) {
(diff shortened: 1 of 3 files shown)


asfgit:

asfgit:

core-job/src/main/java/org/apache/kylin/job/execution/ExecutableManager.java
```diff
@@ -414,6 +414,22 @@ public void pauseJob(String jobId) {
             return;
         }
 
+        if (!(job.getStatus() == ExecutableState.READY
+            || job.getStatus() == ExecutableState.RUNNING)) {
+            logger.warn("The status of job " + jobId + " is " + job.getStatus().toString()
```
nichunen: why not throw an exception

guangxuCheng: Add IllegalStateException

PR #586, opened 2019-04-08
KYLIN-3608 Move dependency versions to top level pom properties

weiqxu:


asfgit:

asfgit:

PR #602, opened 2019-04-14
KYLIN-3939 add BOM character,slove the bug that it shows Chinese garb…

zhaojintaozhao:
User can query sql in kylin portal and  then press "Export" button to download the result; the result is  csv files. User can open the csv file using excel to get the query results.
 But I find that if the result has Chinese characters, the excel will show Chinese characters garble. 

I find that if i open csv file using "notepad++", the Chinese characters are displayed normally. Then I convert this file to "UTF-8-BOM", then save this file. Then I open csv file using excel and the Chinese characters are displayed normally.
Then I add BOM identifier to the query result to be output in "QueryController.java"，I can open csv files usring excel and it displays normally. The system work fine after this change.
This commit just fix this bug.

### DIFF
--- a/server-base/src/main/java/org/apache/kylin/rest/controller/QueryController.java
+++ b/server-base/src/main/java/org/apache/kylin/rest/controller/QueryController.java
@@ -163,7 +163,13 @@ public class QueryController extends BasicController {
             // KYLIN-3939
             // Add BOM character,slove the bug that it shows Chinese garbled when using
             // excel to open scv file on windows.
-            csvWriter.write(BOM_CHARACTER);
+            // BOM character should add on head of CSV file.
+            // So add it to the head of the first index of headerList.
+            if(headerList.size()>0){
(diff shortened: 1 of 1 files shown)


asfgit:

zhaojintaozhao:
If the result of query has Chinese character, then press "Export" button it will download a csv file. Then if you open this file using excel in windows it will show Chinese garble. This PR fix this bug.

nichunen:
Tested, there will be an empty line at first in the exported csv file

zhaojintaozhao:
> Tested, there will be an empty line at first in the exported csv file

Hi Chunen

Thanks for your response, I also find the same problem. I add a new commit to resolve the empty line problem.
Best regards.

15036900300:
User can query sql in kylin portal and then press "Export" button to download the result;When I click the "Export" button,I found sessions missing and prompted me to login. So what should I do?

PR #609, opened 2019-04-16
KYLIN-3959 shutdown query cache for realtime olap

hit-lacus:


asfgit:

PR #610, opened 2019-04-17
KYLIN-3960, Only update user when login in LDAP environment

Aaaaaaron:


asfgit:

PR #612, opened 2019-04-17
KYLIN-3961 Optimize TopNCounter's merge function to reduce TopNCounter's error size.

zhaojintaozhao:
Sometimes TopN measure query will return a large error.
I optimize TopNCounter's merge function to reduce TopNCounter's errors when using TopN measure.
This optimization work well in my kylin system and reduce TopNCounter's error size.

### DIFF
--- a/core-metadata/src/main/java/org/apache/kylin/measure/topn/TopNCounter.java
+++ b/core-metadata/src/main/java/org/apache/kylin/measure/topn/TopNCounter.java
@@ -155,25 +155,12 @@ public class TopNCounter<T> implements Iterable<Counter<T>>, java.io.Serializabl
      * @return
      */
     public TopNCounter<T> merge(TopNCounter<T> another) {
-        boolean thisFull = this.size() >= this.capacity;
-        boolean anotherFull = another.size() >= another.capacity;
-        double m1 = thisFull ? this.counterList.getLast().count : 0.0;
-        double m2 = anotherFull ? another.counterList.getLast().count : 0.0;
-
-        if (anotherFull == true) {
-            for (Counter<T> entry : this.counterMap.values()) {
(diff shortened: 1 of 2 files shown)


asfgit:

zhaojintaozhao:
> Hi jintao, I understand your change.
> 
> The current algorithm bases the algorithm of "a parallel space saving algorithm for frequent items and the Hurwitz zeta distribution", which you can find the link from the reference list of:
> 
> https://kylin.apache.org/blog/2016/03/19/approximate-topn-measure/
> 
> It will make the aggregated value bigger than the actual value, but it is to ensure the position be relatively close to actual.

Hi shaofeng:

I built 3 cubes and test the query performance and accuracy of TOPN.  
The first is normal cube without topN;  The second cube has topN measure with current version code. The third cube has topN measure with my modified code. Both  the topN measure of  the second and third cube is top100. The amount of source data is 22million, I build cube  30 days and using the same query sql. 
I find that both the second  and  third topN cube have a few errors to the actual. But  the second and third topN cube  is very fast than the first normal cube. The error of the  third cube is less than the second cube about position relatively to the actual. The aggregated value of the  second cube is larger than the actual, but the third cube doesn't have error about aggregated value.
Overall, the third modified cube is better than the second current code cube.

nichunen:
Hi @zhaojintaozhao 

I agree that your pr will increase the accuracy of the counter number in some cases, but I'm afraid it will also break the results' oder. Look at the ut I've added in your pr.

The counter with item "DD" is supposed to be the top 1 after two times of merging, but with the code change, it will be  lost

nichunen:
Close this pr for it violates the design of the topn algorithm.
@zhaojintaozhao Please reopen it if you can provide more convincing evidence

PR #614, opened 2019-04-22
Fixed use customized precision

majic31:
In the cubeMeasures.js, It will withdraw precision and scale by using Regular Expression. The scale parameter is ok, but precision use the magic number 19. 
So we fixed it, and we test ok when the column is decimal(38,18).


asfgit:

nichunen:
@majic31 Hi, thanks for contributing. Could you open a jira issue for this?  By the way, commit message in Chinese is not recommended

majic31:
@nichunen Thank you for reply. It's my first pull request in github, so I make some mistakes about the routines, sorry~
So should I open a jira issue and pull request again? (I've never use jira before, would you like to give me some help about using it?)

majic31:
I create issue https://issues.apache.org/jira/browse/KYLIN-3968.

PR #619, opened 2019-04-24
#KYLIN-3977

ZhengshuaiPENG:
1. add ResourceStore.delete(respath, timestamp) interface & impl
2. add ResourceStore.updateTimestamp(respath, timestamp) interface & impl
3. update resource timestamp when reusing dict/snapshot
4. Retry avaibility build process of dict & snapshot 
5. refactor metadata cleanup by using delete with timestamp interface


asfgit:

PR #621, opened 2019-04-26
KYLIN-3987 Set a larger value of reducer num for fact distinct job

Wayne1c:


asfgit:

PR #623, opened 2019-04-28
KYLIN-3968 Customized precision doesn't work in web

majic31:
In the cubeMeasures.js, It will withdraw precision and scale by using Regular Expression. The scale parameter is ok, but precision use the magic number 19.
So we fixed it, In cubeMeasures.js, around line 469:

“var precision = 19;”  --> "var precision = returnValue[2] || 0;"

and we test ok including building cube and querying when the column is decimal(38,18).


asfgit:

PR #627, opened 2019-05-04
replace "\n" by System.lineseperator to support run test case

Qsimple:
replace "\n" by System.lineseperator to support run test case
successfully on windows
platform


asfgit:

PR #632, opened 2019-05-07
KYLIN-3994: StorageCleanupJob may delete cube id data of new built segment because of cube cache in CubeManager

lshmouse:
See: https://issues.apache.org/jira/browse/KYLIN-3994


asfgit:

PR #633, opened 2019-05-07
KYLIN-3994: StorageCleanupJob may delete cube id data of new built segment because of cube cache in CubeManager

lshmouse:
See: https://issues.apache.org/jira/browse/KYLIN-3994

### DIFF
--- a/core-cube/src/main/java/org/apache/kylin/cube/CubeManager.java
+++ b/core-cube/src/main/java/org/apache/kylin/cube/CubeManager.java
@@ -182,14 +182,25 @@ public class CubeManager implements IRealizationProvider {
         }
     }
 
+    /**
+     * List all cubes from cache. Note the metadata may be out of date
+     * @return
+     */
     public List<CubeInstance> listAllCubes() {
         try (AutoLock lock = cubeMapLock.lockForRead()) {
             return new ArrayList<CubeInstance>(cubeMap.values());
         }
     }
 
+    /**
+     * Reload the cubes from database and list all cubes
+     * @return
+     * @throws IOException
+     */
     public List<CubeInstance> reloadAndListAllCubes() throws IOException {
-        crud.reloadAll();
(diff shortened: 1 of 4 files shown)


asfgit:

server-base/src/main/java/org/apache/kylin/rest/job/MetadataCleanupJob.java
```diff
@@ -122,7 +122,7 @@ public MetadataCleanupJob(KylinConfig config) {
 
         // exclude resources in use
         Set<String> activeResources = Sets.newHashSet();
-        for (CubeInstance cube : cubeManager.listAllCubes()) {
+        for (CubeInstance cube : cubeManager.reloadAndListAllCubes()) {
```
ZhengshuaiPENG: In fact here when getting the `toDeleteCandidates`, it filters at candidates who were generated before NEW_RESOURCE_THRESHOLD_MS (12 hours). I think don't need to reload cube here. But it's OK

core-cube/src/main/java/org/apache/kylin/cube/CubeManager.java
```diff
@@ -188,6 +188,11 @@ public void onEntityChange(Broadcaster broadcaster, String entity, Event event,
         }
     }
 
+    public List<CubeInstance> reloadAndListAllCubes() throws IOException {
+        crud.reloadAll();
```
ZhengshuaiPENG: It's better to add lock here to avoid W/R conflict

server-base/src/main/java/org/apache/kylin/rest/job/MetadataCleanupJob.java
lshmouse: In extremely case, NEW_RESOURCE_THRESHOLD_MS (12 hours) may be not enough. So it's better to keep it.

ZhengshuaiPENG:
it's fine to me

PR #634, opened 2019-05-07
KYLIN-3995 config data type may make oom

chenzhx:
jira : https://issues.apache.org/jira/browse/KYLIN-3995


asfgit:

PR #635, opened 2019-05-08
KYLIN-3934 add config for sqoop config null-string and null-non-string

nichunen:


asfgit:

PR #636, opened 2019-05-09
Modify the display format of 'Get job list' api

GinaZhai:


asfgit:

PR #629, opened 2019-05-05
KYLIN-3983 Add extra metadata for measure. Add test case and example/…

Qsimple:
…xample/test_case_data for MeasureManager

UT pass
IT pass

PR #638, opened 2019-05-13
KYLIN-3983 Add extra metadata for measure. Add test case and example/…

Qsimple:
…test_case_data for MeasureManager. Add measure metadata in sample cube

UT passed

IT passed

rewrite some logic according to the advices from https://github.com/apache/kylin/pull/629


asfgit:

PR #647, opened 2019-05-19
KYLIN-2620 TopN can't match when multi sort columns

Wayne1c:


asfgit:

PR #648, opened 2019-05-19
KYLIN-3958 MrHive-Dict support build by livy

hit-lacus:


asfgit:

hit-lacus:
Regression test and manual test passed.

PR #649, opened 2019-05-20
Fixed some look not friendly things

elkan1788:
1. There had found a warning info tip need run as root user when use lastest  version of `Apache Kylin` when check runtime environment.

2. By the way there had found some not friendly variable names and methods in source code,  you can check it with my  git commit.


asfgit:

PR #652, opened 2019-05-21
Removed duplicate code in KylinMeta

TheRealHaui:
Removed duplicate code in KylinMeta


asfgit:

PR #656, opened 2019-05-23
Bug Fix: adjust HBase tableName toUpperCase logic

liuzx32:
**Description**
HBase is case sensitive, tableName.toUpperCase() will lead to the created table can not be found.

**Improvement**
Just remove the toUppderCase() in CubeHFileJob.


asfgit:

PR #657, opened 2019-05-23
adjust HBase tableName toUpperCase for HFile

liuzx32:
### Description
HBase is case sensitive, tableName.toUpperCase() will lead to the created table can not be found.

### Improvement
Just remove the toUppderCase() in CubeHFileJob.


asfgit:

nichunen:
> ### Description
> HBase is case sensitive, tableName.toUpperCase() will lead to the created table can not be found.
> 
> ### Improvement
> Just remove the toUppderCase() in CubeHFileJob.

@liuzx32 Hi, why the upper case of Hbase table name bother you? Can you describe the case of "table can not be found" in detail?

liuzx32:
@nichunen Because HBase tableName is case sensitive. HBase tableName.toUpperCase() will different from the original table, leading to table can not be found.

PR #660, opened 2019-05-26
KYLIN-3983 Add extra metadata for measure. Fix the dead lock cause by…

Qsimple:
dead lock bugfix


asfgit:

PR #663, opened 2019-05-26
KYLIN-3845 solve the problem that kylin build error in stream build task.

zhaojintaozhao:
KYLIN-3845 solve the problem that kylin build error if Kafka streaming source  lacks selected dimensions or metrics then throws ArrayINdexOutofBoundsException  in a kylin stream build task,  .


asfgit:

PR #664, opened 2019-05-28
KYLIN-4017 Build engine get zk(zookeeper) lock failed when building job, it causes the whole build engine doesn't work

wangxiaojing123:
```
【Type】：BUG 
【Severity】：1-Blocker
【Module】：Build Engine
【Description】：Kylin build engine occasionally appears to be unable to get the ZK lock exception, and once this build engine appears, it will not work and can only be restarted to solve.Usually this problem will recur one day after the build engine starts.
【Design】：Setting the cache for curator is never invalid (unless the service stops) and check the state before use curator instance (if closed ,create a new curator instance and put into the cache)
```
```
issue: https://issues.apache.org/jira/browse/KYLIN-4017
```

### DIFF
--- a/core-common/src/main/java/org/apache/kylin/common/util/ZKUtil.java
+++ b/core-common/src/main/java/org/apache/kylin/common/util/ZKUtil.java
@@ -84,7 +84,7 @@ public class ZKUtil {
                         logger.error("Error at closing " + curator, ex);
                     }
                 }
-            }).expireAfterWrite(1, TimeUnit.DAYS).build();
+            }).expireAfterWrite(10000, TimeUnit.DAYS).build();//never expired
 
     static {
         Runtime.getRuntime().addShutdownHook(new Thread(new Runnable() {
--- a/core-job/src/main/java/org/apache/kylin/job/lock/zookeeper/ZookeeperDistributedLock.java
+++ b/core-job/src/main/java/org/apache/kylin/job/lock/zookeeper/ZookeeperDistributedLock.java
@@ -24,6 +24,7 @@ import java.util.Random;
 import java.util.concurrent.Executor;
 
 import org.apache.curator.framework.CuratorFramework;
+import org.apache.curator.framework.imps.CuratorFrameworkState;
 import org.apache.curator.framework.recipes.cache.PathChildrenCache;
 import org.apache.curator.framework.recipes.cache.PathChildrenCacheEvent;
 import org.apache.curator.framework.recipes.cache.PathChildrenCacheListener;
@@ -69,7 +70,7 @@ public class ZookeeperDistributedLock implements DistributedLock, JobLock {
 
     // ============================================================================
 
-    final CuratorFramework curator;
+    private CuratorFramework curator;
     final String client;
     final byte[] clientBytes;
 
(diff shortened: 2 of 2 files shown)


asfgit:

core-common/src/main/java/org/apache/kylin/common/util/ZKUtil.java
```diff
@@ -84,7 +84,7 @@ public void onRemoval(RemovalNotification<String, CuratorFramework> notification
                         logger.error("Error at closing " + curator, ex);
                     }
                 }
-            }).expireAfterWrite(1, TimeUnit.DAYS).build();
+            }).expireAfterWrite(10000, TimeUnit.DAYS).build();//never expired
```
nichunen: Hi, what's the purpose of making the cache of curator client never expire?

wangxiaojing123: > if the cache expire after 1 day,then will run  curator.close(),in other words the newZookeeperClient will closed, but the newZookeeperClient should be as start state all the build engine lifecycle ,it used when build segment.if newZookeeperClient.state!=start,it can't get zk lock ,can't build ：

DistributedScheduler

```java
public void run() {
            try (SetThreadName ignored = new SetThreadName("Scheduler %s Job %s",
                    System.identityHashCode(DistributedScheduler.this), executable.getId())) {
                if (jobLock.lock(getLockPath(executable.getId()))) {
                    logger.info(executable.toString() + " scheduled in server: " + serverName);

                    context.addRunningJob(executable);
                    jobWithLocks.add(executable.getId());
                    executable.execute(context);
                }
            } catch (ExecuteException e) {
                logger.error("ExecuteException job:" + executable.getId() + " in server: " + serverName, e);
            } catch (Exception e) {
                logger.error("unknown error execute job:" + executable.getId() + " in server: " + serverName, e);
            } finally {
                context.removeRunningJob(executable);
                releaseJobLock(executable);
                // trigger the next step asap
                fetcherPool.schedule(fetcher, 0, TimeUnit.SECONDS);
            }
        }
```


 ZookeeperDistributedLock:
 ```java
public boolean lock(String lockPath) {
        logger.debug("{} trying to lock {}", client, lockPath);
        try {
            curator.create().creatingParentsIfNeeded().withMode(CreateMode.EPHEMERAL).forPath(lockPath, clientBytes);
        } catch (KeeperException.NodeExistsException ex) {
            logger.debug("{} see {} is already locked", client, lockPath);
        } catch (Exception ex) {
            throw new IllegalStateException("Error while " + client + " trying to lock " + lockPath, ex);
        }

        String lockOwner = peekLock(lockPath);
        if (client.equals(lockOwner)) {
            logger.info("{} acquired lock at {}", client, lockPath);
            return true;
        } else {
            logger.debug("{} failed to acquire lock at {}, which is held by {}", client, lockPath, lockOwner);
            return false;
        }
    }
```

nichunen: Thanks for your reply, I understand your comment. My concern is why after the cache of curator expired, the newly created zookeeper client is not in state of CuratorFrameworkState.STARTED?

wangxiaojing123: @nichunen Hello nichunen,Thank you reply. "My concern is why after the cache of curator expired, the newly created zookeeper client is not in state of CuratorFrameworkState.STARTED?" Besause now  the code never start a new zookeeper client, it use the old one which state is closed。The new code in my patch I add two changes:1、never expire,2、in case of client is closed, create a new one (but before ,there is no this feature) .

nichunen: OK

PR #666, opened 2019-05-29
KYLIN-4020 fix_length rowkey encode without sepecified length can be …

Qsimple:
…saved but cause CreateHTable step failed


asfgit:

PR #667, opened 2019-05-30
KYLIN-4032 return timestamp and date without timezone in jdbc

chenzhx:


asfgit:

PR #670, opened 2019-05-31
KYLIN-4025: Add detail exception in kylin http response

lshmouse:
See: https://issues.apache.org/jira/browse/KYLIN-4025


asfgit:

PR #671, opened 2019-05-31
KYLIN-4026: Avoid too many file append operations in HiveProducer of of hive metrics reporter

lshmouse:
See: https://issues.apache.org/jira/browse/KYLIN-4026


asfgit:

PR #672, opened 2019-06-03
KYLIN-4028 Speed up startup progress using cached dependency

TempleZhou:


asfgit:

nichunen:
Tested, works fine.

PR #676, opened 2019-06-05
KYLIN-4033 Can not access Kerberized Cluster with DebugTomcat

TempleZhou:


asfgit:

server/src/main/java/org/apache/kylin/rest/DebugTomcat.java
```diff
@@ -99,6 +107,25 @@ private static File findFile(String dir, String ptn) {
         }
         return null;
     }
+    public static boolean isKerberosEnabled() {
+        Configuration conf = new Configuration();
+        boolean authEnabled = conf.getBoolean("hadoop.security.authorization", false);
+        String authMethod = conf.get("hadoop.security.authentication");
+        return authEnabled && authMethod.toLowerCase(Locale.ROOT).equals("kerberos");
```
hn5092: UserGroupInformation.isSecurityEnabled()

server/src/main/java/org/apache/kylin/rest/DebugTomcat.java
```diff
@@ -99,6 +107,25 @@ private static File findFile(String dir, String ptn) {
...
+    }
+
+    public static void authKrb5() {
+        System.setProperty("java.security.krb5.conf", "/etc/krb5.conf");
+        Configuration conf = new Configuration();
+        UserGroupInformation.setConfiguration(conf);
```
hn5092: UGI needn't set hadoopconf.org.apache.hadoop.security.UserGroupInformation#isSecurityEnabled will init it.

server/src/main/java/org/apache/kylin/rest/DebugTomcat.java
```diff
@@ -99,6 +107,25 @@ private static File findFile(String dir, String ptn) {
...
+        try {
+            UserGroupInformation.loginUserFromKeytab(
```
hn5092: Use another way? Systemprop or env and note to dev doc.

server/src/main/java/org/apache/kylin/rest/DebugTomcat.java
```diff
@@ -99,6 +107,25 @@ private static File findFile(String dir, String ptn) {
...
     }
+    public static boolean isKerberosEnabled() {
+        Configuration conf = new Configuration();
+        boolean authEnabled = conf.getBoolean("hadoop.security.authorization", false);
+        String authMethod = conf.get("hadoop.security.authentication");
+        return authEnabled && authMethod.toLowerCase(Locale.ROOT).equals("kerberos");
+    }
+
+    public static void authKrb5() {
+        System.setProperty("java.security.krb5.conf", "/etc/krb5.conf");
```
hn5092: System prop should note to dev doc.

TempleZhou: You are right. And, maybe, the keytab principal and path should be set using System.setProperty or Kylin Config. Do you have any idea?

server/src/main/java/org/apache/kylin/rest/DebugTomcat.java
TempleZhou: Awesome!

PR #680, opened 2019-06-10
KYLIN-4035 Calculate column cardinality by using spark engine

majic31:
link to https://issues.apache.org/jira/browse/KYLIN-4035
Support calculating column cardinality by using spark engine

### DIFF
--- a/core-common/src/main/java/org/apache/kylin/common/KylinConfigBase.java
+++ b/core-common/src/main/java/org/apache/kylin/common/KylinConfigBase.java
@@ -1430,6 +1430,10 @@ public abstract class KylinConfigBase implements Serializable {
         return Boolean.parseBoolean(getOptional("kylin.engine.spark-fact-distinct", "false"));
     }
 
+    public boolean isSparkCardinalityEnabled(){
+        return Boolean.parseBoolean(getOptional("kylin.engine.spark-cardinality", "false"));
+    }
+
     // ============================================================================
     // ENGINE.LIVY
     // ============================================================================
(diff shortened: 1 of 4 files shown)


asfgit:

asfgit:

engine-spark/src/main/java/org/apache/kylin/engine/spark/SparkColumnCardinality.java
```diff
@@ -0,0 +1,158 @@
...
+                    .reduceByKey((x, y) -> {
+                        x.merge(y);
+                        return x;
+                    })
+                    .mapToPair(record -> {
+                        return new Tuple2<>(record._1, record._2.getCountEstimate());
+                    })
+                    .sortByKey(true, 1);
+
+            if (resultRdd.count() == 0) {
```
hit-lacus: Both count and saveAsNewAPIHadoopFile are action of RDD, I think here should be cached to avoid recompute, am I right?

hit-lacus:
In my hadoop cluster(JDK8, hadoop2.6-cdh5.6 with spark-2.3.3-hadoop2.6), I have verfied this patch pass happy  path with correct result.

### By MR
image
### By Spark
image

### Yarn Successful Jobs
image

core-common/src/main/java/org/apache/kylin/common/KylinConfigBase.java
```diff
@@ -1430,6 +1430,10 @@ public boolean isSparkFactDistinctEnable() {
         return Boolean.parseBoolean(getOptional("kylin.engine.spark-fact-distinct", "false"));
     }
 
+    public boolean isSparkCardinalityEnabled(){
+        return Boolean.parseBoolean(getOptional("kylin.engin.spark-cardinality", "false"));
```
hit-lacus: "engin" should be "engine"

PR #689, opened 2019-06-17
KYLIN-4047 Use push-down query when division dynamic column cube query is not supported

wangxiaojing123:
Scene：

SQL:select round (car_dismatch_cnt*100/answer_order_cnt, 4) from db.table1.

- 1、Executes the SQL on a  project where its cube has db.table1,it will throws the right side of the BinaryTuple Expressions columns is not supported for / exception and is not queried through pushdown;

- 2、When change a project where not create the table cube , the same query  can be executed to push down and get the right result.

- 3、 Expect that no matter whether a table cube is created or not, when the cube query is not success, the result can still be obtained by pushdown query, which will give users a consistent experience.


asfgit:

asfgit:

asfgit:

asfgit:

PR #704, opened 2019-06-21
Minor, add Hive JDBC dependency for DebugTomcat

TempleZhou:
Cubing job will failed with kylin.source.hive.client=beeline when I launch Kylin Server locally.


asfgit:

PR #708, opened 2019-06-23
KYLIN-4031 RestClient will throw exception with message contains clea…

Qsimple:
…r-text password


asfgit:

PR #709, opened 2019-06-23
KYLIN-3982 Add measures without purging segments

Qsimple:


asfgit:

PR #715, opened 2019-06-25
KYLIN-4055 cube quey and ad-hoc query return different meta info

javalife0312:
when cube quey then jdbc meta is upper
when ad-hoc quey of presto then jdbc meta is lower

ResultSetMetaData::getColumnLabel
ResultSetMetaData::getColumnName

### DIFF
--- a/query/src/main/java/org/apache/kylin/query/adhoc/PushDownRunnerJdbcImpl.java
+++ b/query/src/main/java/org/apache/kylin/query/adhoc/PushDownRunnerJdbcImpl.java
@@ -214,6 +214,9 @@ public class PushDownRunnerJdbcImpl extends AbstractPushdownRunner {
             // Fall back to default value
         }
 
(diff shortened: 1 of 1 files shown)


asfgit:

query/src/main/java/org/apache/kylin/query/adhoc/PushDownRunnerJdbcImpl.java
```diff
@@ -214,6 +214,9 @@ private SelectedColumnMeta extractColumnMeta(ResultSetMetaData resultSetMetaData
             // Fall back to default value
         }
 
+        columnLabel = columnLabel.toUpperCase(Locale.ROOT);
```
nichunen: Why do upper case convert here?
It seems that there is possibility "columnLabel" or "columnName" be null if SQLException caught.

nichunen:
Will be done by https://github.com/apache/kylin/pull/823/

PR #722, opened 2019-06-30
KYLIN-3628 Fix unexpected exception for select from Lookup Table

hit-lacus:
As Vsevolod Ostapenko report in [JIRA](https://issues.apache.org/jira/browse/KYLIN-3628?focusedCommentId=16874997&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16874997), I have reproduce this kind of issue in my env.

In some case, cube returned by findLatestSnapshot didn't contains the snapshot table
for what you need. Because some dimension which exists in model but was removed in cube design phase, query such as "select * from LookupTable" will throw unexpected exception and confused user.


asfgit:

hit-lacus:
Temporarily closed this PR.

PR #734, opened 2019-07-05
KYLIN-4039, fix ZK distributed lock may not release lock when catchin…

ZhengshuaiPENG:
…g interrupt exception


asfgit:

PR #737, opened 2019-07-10
update dockerfile and fix image display error

RupengWang:
- The docker base image is a little bit old and the newest kramdown cannot run properly.
- fix the bug that the 3rd image on the [page]( http://kylin.apache.org/cn/docs/tutorial/cube_build_job.html) cannot display  because of the file extension name.


asfgit:

PR #740, opened 2019-07-15
KYLIN-4010 Improve real-time streaming

hit-lacus:
1. Update event time by timezone, which help to solve fix timezone not align problem
2. Improve code smell

### DIFF
--- a/core-common/src/main/java/org/apache/kylin/common/KylinConfigBase.java
+++ b/core-common/src/main/java/org/apache/kylin/common/KylinConfigBase.java
@@ -2216,6 +2216,13 @@ public abstract class KylinConfigBase implements Serializable {
         return Long.parseLong(getOptional("kylin.stream.metrics.interval", "5"));
     }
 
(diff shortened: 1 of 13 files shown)


hit-lacus:
After discussion with Gang Ma, we agree with current patch need some modification. What we store into storage should be uniform UTC timestamp. Timezone offset should be modified at query server side, not receiver side. 

I will update my patch asap.

wangxiaojing123:
Hello, We have also met this problem at present, which is of great help to us. However, at present, the changed code may only converts the time displayed on the page to the specified time zone, but segment name should still use GMT time, whether this will lead to a record that should have entered A segment, but actually entered B segment. For example, GMT + 8 is currently used, and the user's incoming time MS (2019-09-1015:09), but this record is currently formatted to segment (2019091007-2019091008).Have you ever met this problem?

hit-lacus:
Check https://github.com/apache/kylin/pull/851

PR #742, opened 2019-07-15
#KYLIN-4083, Fact Distinct Column Step may be failed or value lost whe…

ZhengshuaiPENG:
…n hashcode of the UHC column value is Integer.MIN_VALUE


ZhengshuaiPENG:
Closed because another issue fixed it

PR #753, opened 2019-07-18
KYLIN-4092: Support setting seperate jvm params for kylin backgroud tools

lshmouse:
See: https://issues.apache.org/jira/browse/KYLIN-4092


nichunen:
LGTM

PR #757, opened 2019-07-18
KYLIN-3917: Add max segment merge span to cleanup intermediate data of cube building

lshmouse:
See: https://issues.apache.org/jira/browse/KYLIN-3917

### DIFF
--- a/core-metadata/src/main/java/org/apache/kylin/metadata/model/Segments.java
+++ b/core-metadata/src/main/java/org/apache/kylin/metadata/model/Segments.java
@@ -25,6 +25,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 
+import org.apache.kylin.common.KylinConfig;
 import org.apache.kylin.common.util.ClassUtil;
 import org.apache.kylin.common.util.Pair;
 import org.apache.kylin.metadata.model.SegmentRange.Endpoint;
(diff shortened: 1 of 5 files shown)


core-metadata/src/main/java/org/apache/kylin/metadata/model/Segments.java
```diff
@@ -189,6 +195,21 @@ public void removeLatestSegmentByVolatileRange(Segments<T> segs, long volatileRa
 
     }
 
+    public void removeMaxSpanSegment(Segments<T> segs, long maxSegSpan) {
+        if (maxSegSpan <= 0) {
+            return;
+        }
+        Segments maxSpanSegs = new Segments();
+        for (T seg : segs) {
```
nichunen: Is there any possibility of inner segment(s) being skipped? 
For instance, there are 3 segments like
20190101-20190102, 20190102-20190210, 20190110-20190211.
The method will remove 20190102-20190210, and 2 no-adjacent segments try to be merged

lshmouse: @nichunen Added a test for it

nichunen: @lshmouse OK

PR #758, opened 2019-07-18
KYLIN-4095: Add RESOURCE_PATH_PREFIX option in ResourceTool

lshmouse:
See: https://issues.apache.org/jira/browse/KYLIN-4095

### DIFF
--- a/core-common/src/main/java/org/apache/kylin/common/persistence/ResourceTool.java
+++ b/core-common/src/main/java/org/apache/kylin/common/persistence/ResourceTool.java
@@ -51,8 +51,8 @@ public class ResourceTool {
 
         if (args.length == 0) {
             System.out.println("Usage: ResourceTool list  RESOURCE_PATH");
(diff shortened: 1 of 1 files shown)


core-common/src/main/java/org/apache/kylin/common/persistence/ResourceStore.java
```diff
@@ -474,7 +474,7 @@ public Object call() throws IOException {
      * delete a resource, does nothing on a folder
      */
     final public void deleteResource(String resPath) throws IOException {
-        logger.trace("Deleting resource {} (Store {})", resPath, kylinConfig.getMetadataUrl());
+        logger.info("Deleting resource " + resPath + " (Store " + kylinConfig.getMetadataUrl() + ")");
```
nichunen: I doubt there will be too much log of this one

lshmouse: OK. Revert it. Thanks~

PR #759, opened 2019-07-18
KYLIN-4096: Make cube metadata validator rules configuable

lshmouse:
See: https://issues.apache.org/jira/browse/KYLIN-4096

### DIFF
--- a/core-common/src/main/java/org/apache/kylin/common/KylinConfigBase.java
+++ b/core-common/src/main/java/org/apache/kylin/common/KylinConfigBase.java
@@ -2219,4 +2219,8 @@ public abstract class KylinConfigBase implements Serializable {
     public int getStaleJobThresholdInDays() {
         return Integer.parseInt(getOptional("kylin.tool.health-check.stale-job-threshold-days", "30"));
     }
+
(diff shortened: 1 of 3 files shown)


nichunen:
@lshmouse Hi, would you please submit another pr of blog about how to use this feature?

lshmouse:
@nichunen In xiaomi, we implements 3 rules currently:
- ConfOverrideRule. 
   It check if kylin.engine.spark-conf.spark.yarn.queue/kylin.source.hive.config-override.mapreduce.job.queuename/kylin.engine.mr.config-override.mapreduce.job.queuename must be set in cube Config to make different cube building job running on users'q queues.
  It also check if some forbidden config set in cube level like: kylin.storage.partition.max-scan-bytes/kylin.query.max-scan-bytes/kylin.cube.aggrgroup.max-combination. If user set wrong values of these configs, it will affect the stability of kylin services.
- CubeInfoRule. It make sure: Notification Email List must be set in Cube Info
- AdvancedSettingRule: It make sure Retention Threshold and Volatile Range must be set in Advanced Setting.

In my opinion, these rules may be too specialized. If some of these rule are common, i will add it in the PR.

PR #760, opened 2019-07-18
KYLIN-4097: Throw exception when too many dict slice evictions in AppendTrieDictionary

lshmouse:
See: https://issues.apache.org/jira/browse/KYLIN-4097

### DIFF
--- a/core-common/src/main/java/org/apache/kylin/common/KylinConfigBase.java
+++ b/core-common/src/main/java/org/apache/kylin/common/KylinConfigBase.java
@@ -500,6 +500,10 @@ public abstract class KylinConfigBase implements Serializable {
         return Long.parseLong(getOptional("kylin.dictionary.max-cache-entry", "3000"));
     }
 
+    public int getCachedDictMaxSize() {
+        return Integer.parseInt(getOptional("kylin.dictionary.max-cache-size", "-1"));
+    }
+
     public boolean isGrowingDictEnabled() {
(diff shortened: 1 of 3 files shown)


core-common/src/main/java/org/apache/kylin/common/KylinConfigBase.java
```diff
@@ -2219,4 +2219,8 @@ public int getStaleCubeThresholdInDays() {
     public int getStaleJobThresholdInDays() {
         return Integer.parseInt(getOptional("kylin.tool.health-check.stale-job-threshold-days", "30"));
     }
+
+    public int getDictionarySliceEvicationThreshold() {
```
nichunen: This config should be located in the domain of "DICTIONARY & SNAPSHOT"

nichunen:
Hi @lshmouse  Would you please add a ut for this at AppendTrieDictionaryTest?

core-common/src/main/java/org/apache/kylin/common/KylinConfigBase.java
lshmouse: done

lshmouse:
@

> Hi @lshmouse Would you please add a ut for this at AppendTrieDictionaryTest?

OK

lshmouse:
@nichunen  Added a UT for the case. Please help to review the PR. Thanks~

PR #767, opened 2019-07-23
[KYLIN-4106] fix Illegal partition for SelfDefineSortableKey when Extract Fact Table Distinct Columns

langdamao:
do % opration before abs ，more detail：https://issues.apache.org/jira/browse/KYLIN-4106


langdamao:
didn't connect with jira

PR #775, opened 2019-07-26
KYLIN-4115 Always  load kafkaConsumerProperties

chenzhx:


asfgit:

source-kafka/src/main/java/org/apache/kylin/source/kafka/util/KafkaClient.java
```diff
@@ -57,16 +57,16 @@ public static KafkaConsumer getKafkaConsumer(String brokers, String consumerGrou
...
-            for (Map.Entry entry : properties.entrySet()) {
-                props.put(entry.getKey(), entry.getValue());
-            }
-        }
         props.put("bootstrap.servers", brokers);
         props.put("key.deserializer", StringDeserializer.class.getName());
         props.put("value.deserializer", StringDeserializer.class.getName());
         props.put("group.id", consumerGroup);
         props.put("enable.auto.commit", "false");
+        if (properties != null) {
```
Sean-Gu: Maybe in terms of design, for these five properties, we just don't want them to be overridden by .xml or .xml.override.

PR #782, opened 2019-08-01
KYLIN-4113 Remove the surplus allCubes field & optimize the request to get the cube name

ClausClaus:
Hello there! I found that I seem to screw up the original warehouse, I re-forked a code, the package of the request function I got services/cubes.js, and. . . In the code block I used async/await, I really can't think of anything else other than doing this after getting the return value. . . Do you have any opinions to guide you? ?

### DIFF
--- a/webapp/app/js/controllers/cubeSchema.js
+++ b/webapp/app/js/controllers/cubeSchema.js
@@ -18,7 +18,7 @@
 
 'use strict';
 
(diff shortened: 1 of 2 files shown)


luguosheng1314:
https://www.caniuse.com/#search=await 

Merge first, There are other similar situations in the project, and I think it is better to adapt the new grammar by grunt.js configuration.

ClausClaus:
> https://www.caniuse.com/#search=await
> 
> Merge first, There are other similar situations in the project, and I think it is better to adapt the new grammar by grunt.js configuration.

okay thank you! In fact, I recently configured grunt-babel in the Gruntfile.js file to convert the code so that I can use the newer front-end API in the development. I found that there are a lot of errors in the converted code. This feature needs to be improved. . . I recently read the code in the front end of kylin. I feel that the front-end part of the code seems to be older now. If the team has the idea of ​​refactoring the front-end code, I will be happy to get involved and improve it.

PR #786, opened 2019-08-03
KYLIN-4121 Cleanup hive view intermediate tables after job be finished

RupengWang:
Solve the problem that won't drop hive view intermediate table when the job to build cube that with hive view finished.


RupengWang:
CI passed

PR #793, opened 2019-08-10
KYLIN-4127 Remove never used classes

weibin0516:


weibin0516:
Class `org.apache.kylin.stream.core.storage.StreamingCubeSegment.SegmentInfo`
is never used.  We should remove it.

PR #796, opened 2019-08-10
KYLIN-2820 Query can't read window function's result from subquery

nichunen:


asfgit:

PR #799, opened 2019-08-12
KYLIN-4131 Fix Broadcaster memory leak

Wayne1c:


Wayne1c:
CI passed

core-common/src/main/java/org/apache/kylin/common/Closeable.java
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kylin.common;
+
+public interface Closeable {
```
hit-lacus: In my view, I suggest use `java.io.Closeable` instead. This interface looks needless.

Wayne1c: Because there is no IO operation and do not need throw IOException.

PR #804, opened 2019-08-20
KYLIN-4141 Build Global Dictionary in no time

hit-lacus:
Currently, realtime OLAP do not support `COUNT_DISTINCT(bitmap)` because of lack the ability of encoding string at once. Here I want to use RocksDB & HBase as implementation of streaming distributed dictionary to do that.


hit-lacus:
### Known problem

- [ ] Retry logic need to added to `RemoteDictionaryStore` in case of unavailable of HBase cluster.
- [ ] Rocksdb need to be add more specific config.

hit-lacus:
Check https://github.com/apache/kylin/pull/851

PR #812, opened 2019-08-28
KYLIN-4150 Improve docker for kylin instructions

weibin0516:


hit-lacus:
I think this is fine to me.

PR #816, opened 2019-08-30
Fix admin service test error 

RupengWang:
The commit ["Add time filter for current day jobs"](https://github.com/apache/kylin/commit/ebff43a81eeac79429138d2be1ab7a9a0d5f41f3) whose commit id is 
 ebff43a81eeac79429138d2be1ab7a9a0d5f41f3 will cause Test case AdminServiceTest. testGetPublicConfig() error because expected kyiln config is unequal with kylin-default.properties


RupengWang:
Manual CI passed

PR #825, opened 2019-09-04
KYLIN-4157 When using PrepareStatement query, functions within WHERE …

mwu3:
…will cause InternalErrorException


mwu3:
Already filled root cause and solution on [KYLIN-4157](https://issues.apache.org/jira/browse/KYLIN-4157)
Please check that.

PR #835, opened 2019-09-06
KYLIN-4159 The first step of job will fail and throw "Column \'xx\' i…

luguosheng1314:
…n where clause is ambiguous" in jdbc datasource

### DIFF
--- a/source-jdbc/src/main/java/org/apache/kylin/source/jdbc/JdbcHiveInputBase.java
+++ b/source-jdbc/src/main/java/org/apache/kylin/source/jdbc/JdbcHiveInputBase.java
@@ -17,7 +17,18 @@
  */
 package org.apache.kylin.source.jdbc;
 
-import com.google.common.collect.Maps;
+import java.sql.Connection;
+import java.sql.DatabaseMetaData;
+import java.sql.ResultSet;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Set;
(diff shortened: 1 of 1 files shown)


source-jdbc/src/main/java/org/apache/kylin/source/jdbc/JdbcHiveInputBase.java
```diff
@@ -222,11 +223,13 @@ protected AbstractExecutable createSqoopToFlatHiveStep(String jobWorkingDir, Str
 
             String splitTableAlias;
             String splitColumn;
+            String quoteFullNamedColumn; // `table.column`
             String splitDatabase;
             TblColRef splitColRef = determineSplitColumn();
             splitTableAlias = splitColRef.getTableAlias();
 
             splitColumn = getColumnIdentityQuoted(splitColRef, jdbcMetadataDialect, metaMap, true);
+            quoteFullNamedColumn = quoteIdentifier(partCol, jdbcMetadataDialect.getDialect()).toLowerCase();
```
hit-lacus: Use  `toLowerCase` is forbidden.
```
Forbidden method invocation: java.lang.String#toLowerCase() [Uses default locale]
[ERROR]   in org.apache.kylin.source.jdbc.JdbcHiveInputBase$JdbcBaseBatchCubingInputSide (JdbcHiveInputBase.java:232)
```
[ERROR] Scanned 34 (and 354 related) class file(s) for forbidden API invocations (in 0.10s), 1 error(s).

luguosheng1314: fixed

PR #837, opened 2019-09-12
KYLIN-3973: InvalidProtocolBufferException

zhoukangcn:
See: https://issues.apache.org/jira/browse/KYLIN-3973

Notice: This is just a quick fix, not the best solution.  The best way is using shaded protobuf:2.5.0


shaoxiaowei92:
@zhoukangcn  Did you just modify the CubeVisitProtos. Java class? I modified CubeVisitProtos. java based on githup .After compiling and packaging, replace the jar package under lib with a new one. But the query error still appears after restarting kylin service.

shaoxiaowei92:
@zhoukangcn  Thank you very much.  The problem has now been solved for the time being. Looking forward to resolving this bug in the new version of kylin

PR #840, opened 2019-09-16
KYLIN-4162 After the paragraph is dropped, continue to build the subsequent segment and provide REST API to resubmit the segment

zhangayqian:


server-base/src/main/java/org/apache/kylin/rest/controller/CubeController.java
```diff
@@ -855,6 +863,38 @@ public CubeRequest updateCubeDesc(@RequestBody CubeRequest cubeRequest) throws J
...
+            CubingJob cubingJob = (CubingJob) executable;
+            newSeg.setLastBuildJobID(cubingJob.getId());
+
+            SegmentBuildState.BuildState state = new SegmentBuildState.BuildState();
+            state.setBuildStartTime(System.currentTimeMillis());
+            state.setState(SegmentBuildState.BuildState.State.BUILDING);
+            state.setJobId(cubingJob.getId());
+            streamMetadataStore.updateSegmentBuildState(cubeName, segmentName, state);
+        } catch (Exception e) {
+            logger.error("streaming job submit fail, cubeName:" + cubeName + " segment:" + segmentName, e);
```
nichunen: Do not eat up the exception, throw it.

server-base/src/main/java/org/apache/kylin/rest/controller/CubeController.java
```diff
@@ -855,6 +863,38 @@ public CubeRequest updateCubeDesc(@RequestBody CubeRequest cubeRequest) throws J
 
     }
 
+    /**
+     * resubmit streaming segment
+     *
+     * @throws IOException
+     */
+    @RequestMapping(value = "/{cubeName}/segs/{segmentName}/resubmit", method = {RequestMethod.PUT}, produces = {"application/json"})
```
nichunen: How about adding this api as an action item for a streaming cube?

server-base/src/main/java/org/apache/kylin/rest/controller/CubeController.java
zhangayqian: Ok, I got it.

server-base/src/main/java/org/apache/kylin/rest/controller/CubeController.java
zhangayqian: May I need to add a 'resubmit segment' action item to the streaming cube page?

server-base/src/main/java/org/apache/kylin/rest/controller/CubeController.java
```diff
@@ -855,6 +863,40 @@ public CubeRequest updateCubeDesc(@RequestBody CubeRequest cubeRequest) throws J
 
     }
 
+    /**
+     * resubmit streaming segment
+     *
+     * @throws IOException
+     */
+    @RequestMapping(value = "/{cubeName}/segs/{segmentName}/resubmit", method = { RequestMethod.PUT }, produces = {
```
nichunen: Why not put the api in JobController?

PR #844, opened 2019-09-19
KYLIN-4100: call overview function when initing job ctrl in job.js (addendum)

lshmouse:
See: KYLIN-4100


lshmouse:
x

Found the overview api called twice. After searching it, it may be a issue of js but I am not familiar with angular. Can anyone can help? Thanks~

PR #850, opened 2019-09-21
KYLIN-4010 Support date format "yyyy-MM-dd'T'HH:mm:ss.SSSZZ"

nichunen:


hit-lacus:
I think it is good, because user may use timestamp with timezone, we should recognize that pattern. Better add UT.

PR #853, opened 2019-09-23
KYLIN-4176 Filter the intermediate tables when loading table metadata from tree

wkh8011:
Load table metadata from tree in 'Data Source' tab, then choose one database, you will find the intermediate tables are also loaded, But intermediate tables are the inner tables, which should not be showed to client, so I filter them in the kylin server.


wkh8011:
Signed-off-by: wkh8011 <wu.kehua@zte.com.cn>

PR #863, opened 2019-09-29
Add cube info and hbase region name in log where hbase coprocessor deadline happended

SunFlowerY1:
When there is a large number of query, hbase server would be overload and corprocessor deadline would be happended so that there would be a large number of query timeout. We need cube and region infomation to find the problem and recovery query service. But currently there is no cube 
 and region infomation about query in log when corprocessor deadline happeded.We add it to help finding the bad query or the problem of hbase server so that we can recovery query service quickly.


lshmouse:
@luguosheng1314 Could you help to compress the commits of PR into one? Please see:
https://www.internalpointers.com/post/squash-commits-into-one-git

nichunen:
Will be done in https://github.com/apache/kylin/pull/1021, close this one

PR #864, opened 2019-09-29
KYLIN-4181 Schedule Kylin using Kubernetes

TempleZhou:


kubernetes/kylin-statefulset.yaml
```diff
@@ -0,0 +1,78 @@
...
+    metadata:
+      labels:
+        app: kylin
+    spec:
+      containers:
+        - command:
+            - bash
+            - '-c'
+            - '${KYLIN_HOME}/bin/kylin.sh run'
+          image: 'kylin:3.0.0-alpha2-cdh57'
```
lshmouse: Can we config the image name?

kubernetes/Dockerfile
```diff
@@ -0,0 +1,63 @@
+FROM centos:centos7
+
+ENV CDH_VERSION     5.7.6
+ENV JAVA_VERSION    1.8.0
```
lshmouse: A Dockerfile has been committed to Kylin. Could yout reuse it ?

kubernetes/README.md
```diff
@@ -0,0 +1,108 @@
...
+Then, you need to create headless service for stable DNS entries(kylin-0.kylin, kylin-1.kylin, kylin-2.kylin...) of StatefulSet members.
+
+```bash
+kubectl apply -f kylin-service.yaml
+```
+
+Finally, create the StatefulSet and try to use it:
+
+```bash
+kubectl apply -f kylin-statefulset.yaml
```
lshmouse: It's better to have an ingress support

kubernetes/Dockerfile
TempleZhou: Emm... the committed dockerfile is self-contained and work with Apache version instead of CDH version. It seems to prepare for experiment, I think.

weibin0516: Hi, @TempleZhou , what do you hope? We can work to make it usable in production Environment.

sanjulian: Hi @TempleZhou, how about split docker image as 2 part: one is base image, and the other is kylin image.
Kylin docker image base on base image (hadoop-client image) by 'FROM'.
And we can also provide 2 base image: one is apache version, the other is CDH version.
It will benefit users who owns different version of hadoop version for their environment.
What do you think?

kubernetes/kylin-job-statefulset.yaml
```diff
@@ -0,0 +1,95 @@
...
+          lifecycle:
+            postStart:
+              exec:
+                command:
+                  - bash
+                  - '-c'
+                  - |
+                    set -ex
+                    # initialize the keytab
+                    kinit -kt /home/kylin/kylin.keytab kylin
```
sanjulian: Maybe we can create a bootstrap.sh to includes these command, and exec the shell script in here.

sanjulian: Do we have any keytab refresh requirement? For example, refresh keytab every ** minutes.

TempleZhou:
> @TempleZhou How do you solve the problem of unstable ip when pod restart?

Hi @Smallhi , I don't know why you need to care about the pod IP as we access the Kylin instance using domain (Ingress).

kubernetes/Dockerfile
```diff
@@ -0,0 +1,78 @@
+FROM centos:6.9
+
+ARG APACHE_MIRRORS=http://mirrors.aliyun.com
```
zhaoyongjie: Please use the official apache repository

kubernetes/Dockerfile
```diff
@@ -0,0 +1,78 @@
+FROM centos:6.9
```
zhaoyongjie: It is not recommended to use the operating system as the base image, you can use the base image provided by maven. Https://hub.docker.com/_/maven

Smallhi:
@TempleZhou  In our prod env, we also use k8s as we deploy env, we found that, if we define kylin.server.mode = query ,we also need to define kylin.server.cluster-servers= ip1:7070,ip2:7070...
we deploy kylin in a app ,then whole app share a same domain adress, so , if we want differ role of kylin instance ,we can only use pod ip for kylin.server.cluster-servers, however ,for k8s ,if the pod restart, the ip will fly....

TempleZhou:
> @TempleZhou In our prod env, we also use k8s as we deploy env, we found that, if we define kylin.server.mode = query ,we also need to define kylin.server.cluster-servers= ip1:7070,ip2:7070...
> we deploy kylin in a app ,then whole app share a same domain adress, so , if we want differ role of kylin instance ,we can only use pod ip for kylin.server.cluster-servers, however ,for k8s ,if the pod restart, the ip will fly....

You only need to set the kylin.server.host-address=$hostname, and the kylin instance will register the information in Zookeeper, related issue: https://issues.apache.org/jira/browse/KYLIN-3820

kubernetes/kylin-job-statefulset.yaml
```diff
@@ -0,0 +1,95 @@
...
+                    # set the kylin.server.mode
+                    sed "s/kylin\.server\.mode.*/kylin\.server\.mode=all/g" /mnt/kylin-config/kylin.properties > ${KYLIN_HOME}/conf/kylin.properties
+                    sed -i "s/kylin\.server\.host-address.*/kylin\.server\.host-address=`hostname`\.kylin:7070/g" ${KYLIN_HOME}/conf/kylin.properties
```
TempleZhou: @Smallhi Maybe, you need pay attention to my postStart script.

hit-lacus:
Should close this PR, because it is contained in https://github.com/apache/kylin/pull/1182 and commits is been merged, thank you @TempleZhou .

PR #866, opened 2019-09-30
KYLIN-4183 Clicking 'Submit' button is unresponsive, when the segment is not selected.

wkh8011:
When the segment is not selected, "$scope.lookup.select.segments" will be undefined, which will cause this problem.


luguosheng1314:
LGTM

PR #875, opened 2019-10-11
KYLIN-4190 hiveproducer write() function throw exception because hive…

bigxiaochu:
https://issues.apache.org/jira/browse/KYLIN-4190
hiveproducer write() function throw exception because hive mertics table location path prefix is different with defaut fs when hdfs uses router-based federation


bigxiaochu:
This pull request is from bigxiaochu:master, I am not convenient to checkout branch from bigxiaochu:master for the other pull request .
So I close this request, and open a new request from bigxiaochu:KYLIN-4190

PR #877, opened 2019-10-12
KYLIN-4190 hiveproducer write() function throw exception because hive…

bigxiaochu:
https://issues.apache.org/jira/browse/KYLIN-4190
hiveproducer write() function throw exception because hive mertics table location path prefix is different with defaut fs when hdfs uses router-based federation


lshmouse:
@nichunen Could you help to review it? Thanks~

PR #900, opened 2019-10-28
KYLIN-4207: fix exception while use cube metadata extra validators

zhoukangcn:
see: https://issues.apache.org/jira/browse/KYLIN-4207


nichunen:
@lshmouse OK to me too.

PR #903, opened 2019-10-29
KYLIN-4208 RT OLAP kylin.stream.node configure optimization support all receiver can have the same config

wangxiaojing123:
Support three kylin.stream.node config format, ip:prot or port or not set the config.

- if set ip:port , then kylin will set the config ip and port as the currentNode,

- if set port only, then kylin will get the node ip address and set the node ip and port as the currentNode,

- if not set the config ,then kylin will get the node hostname address and set the hostname and defaultPort(7070) as the currentNode.

issue link: https://issues.apache.org/jira/browse/KYLIN-4208

### DIFF
--- a/stream-core/src/main/java/org/apache/kylin/stream/core/util/NodeUtil.java
+++ b/stream-core/src/main/java/org/apache/kylin/stream/core/util/NodeUtil.java
@@ -23,23 +23,39 @@ import org.apache.kylin.stream.core.model.Node;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.net.Inet4Address;
 import java.net.InetAddress;
+import java.net.NetworkInterface;
 import java.net.UnknownHostException;
+import java.util.Enumeration;
 import java.util.Map;
 
 public class NodeUtil {
     private static final Logger logger = LoggerFactory.getLogger(NodeUtil.class);
 
+    /*
+    * support three kylin.stream.node config format
+    * ip:prot or port or not set the config
+    * if set ip:port , then kylin will set the config ip and port as the currentNode,
+    * if set port only, then kylin will get the node ip address and set the node ip and port as the currentNode,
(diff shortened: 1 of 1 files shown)


stream-core/src/main/java/org/apache/kylin/stream/core/util/NodeUtil.java
```diff
@@ -55,4 +71,29 @@ private static String getLocalhostName() {
         }
         return host;
     }
+
+    private static String getLocalHostIp() {
```
hit-lacus: Hi, from the issue https://issues.apache.org/jira/browse/KYLIN-3747,  the author MaGang recommend to use FQDN other than ip to register receiver. So, in my view, `getLocalHostIp` is not necessary, use `getLocalhostName` maybe better.

hit-lacus:
By the way, after introduce https://issues.apache.org/jira/browse/KYLIN-4084, you could use the `sh bin/kylin-port-replace-util.sh set XXX` to change port for each receiver.

stream-core/src/main/java/org/apache/kylin/stream/core/util/NodeUtil.java
wangxiaojing123: Thank you for your reply. I'm in favor of using host name instead of IP.

wangxiaojing123:
Thank you for your reply. Although we can use tools to replace ports, according to the past operation and maintenance experience of clusters, it will be more standardized to config the port in the some config file.
   In addition, with the increase of nodes in the future, the nodes of the same cluster should use the same configuration content, rather than the content of the file of each node is different.

hit-lacus:
> Thank you for your reply. Although we can use tools to replace ports, according to the past operation and maintenance experience of clusters, it will be more standardized to config the port in the some config file.
> In addition, with the increase of nodes in the future, the nodes of the same cluster should use the same configuration content, rather than the content of the file of each node is different.

OK, I agree with you.

hit-lacus:
Looks good at my side.

PR #909, opened 2019-10-30
KYLIN-4225 close hive session manually

zhoukangcn:
see: https://issues.apache.org/jira/browse/KYLIN-4225


zhoukangcn:
@nichunen Could you help to review it? Thanks

PR #910, opened 2019-10-30
KYLIN-4226: Skip current unavailable tables when updating hbase copro…

zhoukangcn:
…cessor

see: https://issues.apache.org/jira/browse/KYLIN-4226


zhoukangcn:
@lshmouse @nichunen Could you help to review this? Thanks~

PR #920, opened 2019-11-05
KYLIN-4237 Return error when execute "explain plan for SQL" to get th…

zhangayqian:
…e execution plan of SQL


nichunen:
@zhangayqian Please add a ut in QueryUtilTest

PR #921, opened 2019-11-05
Fix streaming model broken when kylin.source.hive.database-for-flat-t…

Markz666:
…able set to value other than default.


Markz666:
https://issues.apache.org/jira/browse/KYLIN-4238

nichunen:
image
Tested, fine to me

PR #934, opened 2019-11-12
KYLIN-4250 FechRunnner should skip the job to process other jobs instead of throwing exception when the job section metadata is not found

bigxiaochu:
https://issues.apache.org/jira/browse/KYLIN-4250


bigxiaochu:
conflicts， close first

PR #939, opened 2019-11-13
KYLIN-4241:Dynamically reload user's security configuration with prop…

zhangxiang17:
…erties or yaml config


shaofengshi:
How does this closed? It looks like a good feature.

PR #940, opened 2019-11-13
KYLIN-4254: Export from Insight with CSV format error, when sql contains Chinese

x92zhang:
1. change the default decoding to 'UTF-8' in tomcat's server.xml


RupengWang:
LGTM

PR #945, opened 2019-11-17
KYLIN-4258 Real-time OLAP may return incorrect result for some case

allenma:
1. When real-time query group by minute_start and filter by another column, the aggregate result maybe not correct, the issue has been addressed in KYLIN-4184 , the fix is good, but in  

NoCompressedColumnReader class, the readBuffer bytes is still shared, it is a potential issue, need to be fixed.

2. When filter by a minute_start range, the first minute data may not be correct, it is caused by:https://github.com/apache/kylin/blob/master/stream-core/src/main/java/org/apache/kylin/stream/core/storage/columnar/ColumnarSegmentStoreFilesSearcher.java#L68, the fragment's max time is inclusive, but this method's max time is exclusive, that may skip some fragments for the first minute by mistake.

3. Add some UT to cover these two cases.


allenma:
@nichunen could you help to review this PR?

PR #950, opened 2019-11-18
KYLIN-4260 When using server side PreparedStatement cache, the query …

mawu2:
…result are not match on TopN scenario


nichunen:
Hi, I think a better fix for this issue is to disable adjusting sql digest if the prepared statement cache is enabled.

mawu2:
Hi, I did several tests, found this solution won't work under some cases, and will cause some other issues. So I'll close this PR request for now.

PR #963, opened 2019-11-27
KYLIN-4271 Use configurable certificate to support LDAPs authentication of Kylin

wkh8011:
we add "kylin.security.ldap.connection-truststore" parameter which is set to be value of "javax.net.ssl.trustStore", so we can use configurable certificate to support LDAPs authentication of Kylin.


nichunen:
Hi @wkh8011 
Would you please add a comment for the configuration item in the properties file and update the document named "Secure with LDAP and SSO"?

PR #971, opened 2019-12-03
KYLIN-3237 Not to allow segment deletion by REST API when its cube is…

zhangayqian:
… in READY status


server-base/src/main/java/org/apache/kylin/rest/msg/Message.java
```diff
@@ -65,6 +65,10 @@ public String getDELETE_NOT_READY_SEG() {
         return "Cannot delete segment '%s' as its status is not READY. Discard the on-going job for it.";
     }
 
+    public String getDELETE_SEG_FROM_READY_CUBE() {
```
nichunen: should also be added in CnMessage.java

zhangayqian: OK，I'll do it.

PR #977, opened 2019-12-07
KYLIN-4283: return SUCCEED while IOException happend in GarbageCollec…

zhoukangcn:
…tionStep

see: https://issues.apache.org/jira/browse/KYLIN-4283


shaofengshi:
I agree; +1

PR #985, opened 2019-12-10
KYLIN-4290: Add file lock to kylin startup script to avoid starting multiple instances on one node

yangzheng10:
When we restart the Kylin instance, due to the slow start of kylin, some monitoring scripts fail to detect the surviving kylin, and will also start the kylin instance. As a result, there are multiple kylin instances on a node.
By adding a file lock to the startup script (kylin.sh), it is guaranteed that only one kylin instance will be started when concurrent startup on a node.

### DIFF
--- a/build/bin/kylin.sh
+++ b/build/bin/kylin.sh
@@ -76,6 +76,14 @@ function retrieveStartCommand() {
         fi
     fi
 
+    lockfile=$KYLIN_HOME/LOCK
+    if [ ! -e $lockfile ]; then
+        trap "rm -f $lockfile; exit" INT TERM EXIT
+        touch $lockfile
+    else
+        quit "Kylin is starting, wait for it"
+    fi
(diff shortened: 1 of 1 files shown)


hit-lacus:
Thank you @zenoyang , I have do a simple check, the problem does exists. Your PR looks good.

hit-lacus:
1. Since this is small enchancement, please try to compact all code change into ONE commit.
2. Please change the author and email in your patch to align to your github profile.

hit-lacus:
I have test it, it looks good to me. @nichunen 
1. run kylin.sh in parallel way.
2. test if start failed, will lock be released automatically.

yangzheng10:
> 1. Since this is small enchancement, please try to compact all code change into ONE commit.
> 2. Please change the author and email in your patch to align to your github profile.

done

PR #991, opened 2019-12-14
KYLIN-4255 Display detailed error message when using livy build error

weibin0516:


weibin0516:
before:
image


after:
image

PR #998, opened 2019-12-16
implement new feature in KYLIN-4301

zengruios:
Please see https://issues.apache.org/jira/browse/KYLIN-4301 for detail.


hit-lacus:
1. Will this cause some inconsistent situation? Does removed receiver also be removed in replica set and cube assignment?
2. I knew in 3.0.0, you can remoe receiver in web UI.

zengruios:
OK, I will see it in 3.0.0.

hit-lacus:
Thank you, @zengruios , you are great!

PR #1008, opened 2019-12-19
KYLIN-4227: add segment build callback

zhoukangcn:
see: https://issues.apache.org/jira/browse/KYLIN-4227

### DIFF
--- a/engine-mr/src/main/java/org/apache/kylin/engine/mr/CubingJob.java
+++ b/engine-mr/src/main/java/org/apache/kylin/engine/mr/CubingJob.java
@@ -26,12 +26,19 @@ import java.util.Map;
 import java.util.TimeZone;
 import java.util.regex.Matcher;
 
+import org.apache.http.impl.client.DefaultHttpClient;
+import org.apache.http.params.BasicHttpParams;
+import org.apache.http.params.HttpConnectionParams;
+import org.apache.http.params.HttpParams;
+import org.apache.http.client.methods.HttpGet;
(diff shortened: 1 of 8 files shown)


zhoukangcn:
@nichunen Could you help to review this? Thank you

engine-mr/src/main/java/org/apache/kylin/engine/mr/CubingJob.java
```diff
@@ -257,6 +266,36 @@ void setJobType(String jobType) {
...
+            CubeDesc cubeDesc = CubeDescManager.getInstance(getConfig()).getCubeDesc(cubeName);
+            if (StringUtils.isBlank(cubeDesc.getCubeCallback())) {
+                logger.info("cubeName:" + cubeName + ", callback url is empty");
+                return;
+            }
+            String segmentId = CubingExecutableUtil.getSegmentId(getParams());
+            CubeSegment cubeSegment = CubingExecutableUtil.findSegment(executableContext, cubeName, segmentId);
+            StringBuilder sb = new StringBuilder(cubeDesc.getCubeCallback());
+            SegmentRange.TSRange tsRange = cubeSegment.getTSRange();
+            sb.append("?cube=").append(cubeName)
```
hit-lacus: 1. Hard code the url parm seems not like a good practice(I am not really sure), maybe it should be in more user-defined way.
2. It is a blocking call (synchronous), will it cause build more slow when the endpoint is down?

yabola:
test

PR #1021, opened 2019-12-26
add cube info and hbase region name in log when hbase coprocessor deadline happened

SunFlowerY1:
When there is some cube scan a lot of data on hbase, hbase coprocessor would happen deadline. Cube info and region name in log can help us locate unhealthy query quickly!


SunFlowerY1:
When there is some cube scan a lot of data on hbase, hbase coprocessor would happen deadline. Cube info and region name in log can help us locate unhealthy query quickly!

PR #1022, opened 2019-12-26
htable current-limiting

SunFlowerY1:
For reduce the hbase load, when create htable we set the htable current-limiting and make is configurable. The parameter name for htable current-limiting is kylin.hbase.htable-quota-limit and the default value is 10000


nichunen:
@SunFlowerY1 Hi，would you please open a jira issue for your code change?

PR #1025, opened 2019-12-27
delete /execute_output metadata when deleting a job

ChangjiGuo:
delete both /execute and /execute_output metadata when deleting a job


ChangjiGuo:
I need to modify it a little, because the metadata for each step is also deleted.

PR #1026, opened 2019-12-27
delete /execute_output metadata at the same time when deleting a job

ChangjiGuo:
Delete both /execute and /execute_output metadata when deleting a job.

### DIFF
--- a/core-job/src/main/java/org/apache/kylin/job/dao/ExecutableDao.java
+++ b/core-job/src/main/java/org/apache/kylin/job/dao/ExecutableDao.java
@@ -378,14 +378,31 @@ public class ExecutableDao {
 
     public void deleteJob(String uuid) throws PersistentException {
         try {
+            ExecutablePO executablePO = getJob(uuid);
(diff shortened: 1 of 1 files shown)


nichunen:
@ChangjiGuo 
Hi, I think your point of this PR is right, job's tasks' should also be deleted once the job is deleted (they will be deleted by MetadataCleanupJob). 
But I doubt the code change will also bring instability of the job deleting operation. Because before this change, dropping a job only needs one time of metadata deleting, but after this change, it needs n+1 times(n is the number of tasks). If one task failed to be deleted, an exception will be thrown which indicates the deletion is failed, but actually, the job's metadata has been deleted.

ChangjiGuo:
Hi @nichunen 
I agree with your doubt. Do you think it is necessary to delete /execute_output?

ChangjiGuo:
How about adding a verification in the catch statement?

ChangjiGuo:
Or delete output separately?

PR #1035, opened 2020-01-02
Collect information of NumberFormatException

baiyunsong:
KYLIN-4324：User query returns Unknown error


baiyunsong:
…eption

see: [https://issues.apache.org/jira/browse/KYLIN-4324](url)

PR #1051, opened 2020-01-11
KYLIN-4334 Auto discard all jobs while disable the Real Time Cube

ikids-star:
- issue link : https://issues.apache.org/jira/browse/KYLIN-4334


hit-lacus:
I have tested it, it works well, let's merge it.

PR #1052, opened 2020-01-11
KYLIN-4326 KYLIN support ZSTD codec done!

Smallhi:
- KylinCodec test:
image

# CubeVisitWithCompressionTest: (mearsure with bitmap)
image


nichunen:
@Smallhi Hi, I just pull your pr, it seems the code compiled with errors. You should add the dependencies with versions in the root pom.xml. What's more, there are conflicts with master branch, please resolve them, thanks

Smallhi:
@shaofengshi  Thanks for your reply! 
I plan to rewrite this feature and solve these problems:
1. use third-party lib licenced under  Apache License v2.
2. If we know the original size of bytes array, when we use zstd  or lz4 to decompress inputstream, we need't use try catch cluase to extend Bytebuffer size, which I think is more efficient.

PR #1054, opened 2020-01-12
Fix some insecure issues

nichunen:


server/src/test/java/org/apache/kylin/rest/util/ValidateUtilTest.java
```diff
@@ -48,6 +48,9 @@
 
     @Test
     public void testCheckIdentifiersExists() throws IOException {
+        String project = "/asd/../_asd33b";
+        project = ValidateUtil.convertStringToBeAlphanumericUnderscore(project);
```
hit-lacus: Could we add some assertEqual for this `project`?

nichunen: @hit-lacus This is some code I submit carelessly, I have removed it

core-common/src/main/java/org/apache/kylin/common/persistence/JDBCResourceStore.java
```diff
@@ -44,28 +45,17 @@
 
 public class JDBCResourceStore extends PushdownResourceStore {
 
-    private static Logger logger = LoggerFactory.getLogger(JDBCResourceStore.class);
-
     public static final String JDBC_SCHEME = "jdbc";
-
+    private static final ConcurrentHashMap<String, Object> resourceHashMap = new ConcurrentHashMap<>();
```
hit-lacus: Maybe change the name `resourceHashMap` to something like `lockObjectMap` is more readable.

nichunen: @hit-lacus Good idea

PR #1056, opened 2020-01-13
KYLIN-4336 Global domain dict for MR build engine

wangxiaojing123:
issue link: https://issues.apache.org/jira/browse/KYLIN-4336

### DIFF
--- a/core-cube/src/main/java/org/apache/kylin/cube/model/CubeDesc.java
+++ b/core-cube/src/main/java/org/apache/kylin/cube/model/CubeDesc.java
@@ -40,6 +40,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.Set;
 import java.util.TreeSet;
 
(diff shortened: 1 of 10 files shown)


core-cube/src/main/java/org/apache/kylin/cube/model/CubeDesc.java
```diff
@@ -1375,12 +1376,37 @@ public void setParentForward(int parentForward) {
                     result.remove(dictDesc.getColumnRef());
                     result.add(dictDesc.getResuseColumnRef());
                 }
+
+                //tiretree global domain dic
+                if (Objects.isNull(dictDesc.getResuseColumnRef()) && Objects.nonNull(dictDesc.getReuseColumn())) {
+                    logger.info("tiretree global domain dic : column {} use tiretree global domain dic, reuse column {} ", dictDesc.getColumnRef(), dictDesc.getReuseColumn());
```
nichunen: I think it's better to put this into the scope of the previous conditional expression "if (dictDesc.getResuseColumnRef() != null) "

core-cube/src/main/java/org/apache/kylin/cube/model/CubeDescTiretreeGlobalDomainDictUtil.java
```diff
@@ -0,0 +1,164 @@
...
+     * add resuce global tiretree global dic for baseid job
+     * @param cubeDesc
+     * @param dumpList
+     */
+    public static void cuboidJob(CubeDesc cubeDesc, Set<String> dumpList) {
+        logger.info("cube {} start to add global domain dic", cubeDesc.getName());
+        CubeManager cubeManager = CubeManager.getInstance(KylinConfig.getInstanceFromEnv());
+        DataModelManager metadataManager =DataModelManager.getInstance(KylinConfig.getInstanceFromEnv());
+
+        cubeManager.getCube(cubeDesc.getName());
```
nichunen: useless

core-cube/src/main/java/org/apache/kylin/cube/model/CubeDescTiretreeGlobalDomainDictUtil.java
```diff
@@ -0,0 +1,164 @@
...
+
+
+
+
+    /**
+     * add resuce global tiretree global dic for baseid job
+     * @param cubeDesc
+     * @param dumpList
+     */
+    public static void cuboidJob(CubeDesc cubeDesc, Set<String> dumpList) {
```
nichunen: The method name is not clear enough to clarify its function

core-cube/src/main/java/org/apache/kylin/cube/model/CubeDesc.java
wangxiaojing123: Thanks for coding review .The previous conditional expression use for reuse col of the same cube (Existing functions of kylin), and dictDesc.getResuseColumnRef() is not null, but the new global domain dic's dictDesc.getResuseColumnRef() is null . After the expression is satisfied, the content to be processed is different too， the global domain dic only result.remove(dictDesc.getColumnRef()) no result.add(dictDesc.getResuseColumnRef()).

core-cube/src/main/java/org/apache/kylin/cube/model/CubeDescTiretreeGlobalDomainDictUtil.java
wangxiaojing123: Thanks ,I will fix it.

core-cube/src/main/java/org/apache/kylin/cube/model/CubeDescTiretreeGlobalDomainDictUtil.java
wangxiaojing123: OK, I will fix it, thanks.

nichunen:
Merged in PR https://github.com/apache/kylin/pull/1058

PR #1059, opened 2020-01-14
KYLIN-4291 Parallel segment building may causes WriteConflictException

zhangayqian:


core-dictionary/src/main/java/org/apache/kylin/dict/DictionaryManager.java
```diff
@@ -193,12 +194,24 @@ private DictionaryInfo updateExistingDictLastModifiedTime(String dictPath) throw
...
         if (StringUtils.isBlank(dictPath))
             return NONE_INDICATOR;
-        long now = System.currentTimeMillis();
-        store.updateTimestamp(dictPath, now);
-        logger.info("Update dictionary {} lastModifiedTime to {}", dictPath, now);
-        DictionaryInfo dictInfo = load(dictPath, true);
-        updateDictCache(dictInfo);
-        return dictInfo;
+
+        int retry = 7;
```
zhoukangcn: I think change constant to KylinConfig is a better way.

zhangayqian: Great advice! Thank you!

PR #1072, opened 2020-01-19
KYLIN-4356: clean outdate column cardinality job

zhoukangcn:
see: https://issues.apache.org/jira/browse/KYLIN-4356


nichunen:
Will this forcedly delete all jobs with types that are not CubingJob? If so, I doubt it's not a good way.

zhoukangcn:
@nichunen code updated, please help to review it.

PR #1073, opened 2020-01-19
KYLIN-4357: throw Exception when cannot find job

zhoukangcn:
see: https://issues.apache.org/jira/browse/KYLIN-4357


nichunen:
Revoke my approve

nichunen:
If a user has a workflow, check whether a job is deleted by calling /job/{uuid}. It will assume a job is deleted if he gets the result "null". The code change will break such workflows.

zhoukangcn:
> If a user has a workflow, check whether a job is deleted by calling /job/{uuid}. It will assume a job is deleted if he gets the result "null". The code change will break such workflows.

Yes, you are right. 
code updated, throw exception just when use null jobInstance.
Thank you!

PR #1074, opened 2020-01-20
KYLIN-4340 Fix bug of get value of isSparkFactDistinctEnable for cube not correct

weibin0516:


nichunen:
@weibin0516 Hi, I have reviewed your pr. But I think it's not a good way to fetch the cube's config in this way. In method org.apache.kylin.engine.spark.SparkBatchCubingJobBuilder2#build, try use "seg.getConfig().isSparkFactDistinctEnable()" instead, I think this should work.

weibin0516:
@nichunen , config isSparkFactDistinctEnable is just a example, this is a bug, this bug will cause all cube config not effective

nichunen (approved):
LGTM

PR #1075, opened 2020-01-20
KYLIN-4358:Bugfix statement cache eviction invalidation base on time

zhaomoran:
`minEvictableIdleTimeMillis ` param is invalid, because `timeBetweenEvictionRunsMillis ` parameter must be used at the same time. [GenericObjectPool](https://commons.apache.org/proper/commons-pool/api-1.6/org/apache/commons/pool/impl/GenericObjectPool.html)
```
timeBetweenEvictionRunsMillis indicates how long the eviction thread should sleep before "runs" of examining idle objects. When non-positive, no eviction thread will be launched. The default setting for this parameter is -1 (i.e., idle object eviction is disabled by default).
minEvictableIdleTimeMillis specifies the minimum amount of time that an object may sit idle in the pool before it is eligible for eviction due to idle time. When non-positive, no object will be dropped from the pool due to idle time alone. This setting has no effect unless timeBetweenEvictionRunsMillis > 0. The default setting for this parameter is 30 minutes.
```


zhaomoran:
https://issues.apache.org/jira/browse/KYLIN-4358

PR #1082, opened 2020-01-27
Kylin 4332

Sidonet:
When feature "add from table tree" added, schema + _ + tableName used in list, due to AngularJS components doesn't know who is the parent of the child.

Add this logic to Reload and Unload table buttons.


nichunen:
@Sidonet Hi, 
Would you please describe the detail steps for reproducing the issue?
It works fine on me when reloading tables on "Data Source".

Sidonet:
Hello, 

Attached to Jira.

Sidonet:
Need more time to investigation.

PR #1108, opened 2020-02-15
Fix sonar issues

etherge:
Fixed 30 sonar issues.
- Fixed 18 issues: Math operands should be cast before assignment
- Fixed 7 issues: Iterator.next() methods should throw NoSuchElementException
- Fixed 1 issue: toString() should not return null
- Fixed 2 issues: Non-thread-safe fields should not be static
- Fixed 1 issue:  compareTo should not be overloaded
- Fixed 1 issue: Externalizable classes should have no-arguments constructors


etherge:
retest this please

PR #1117, opened 2020-02-22
htable name can not cast to upper

fengpod:
htable name can not be casted to upper


shaofengshi:
hello, can you add a description for this pr?

fengpod:
> hello, can you add a description for this pr?

In class CubeHFileJob,htable name be casted to upper on 103th lines，cause TableNotFoundException。as shown in the picture：
image

shaofengshi:
I see; sometimes the hbase namespace is not all in uppercase, so this will cause problem.

PR #1121, opened 2020-02-25
KYLIN-4355 Add validation for cube assignment

hit-lacus:


hit-lacus:
Look like jenkins build failed is caused by env, not CL.

PR #1134, opened 2020-03-02
KYLIN-4396 Close FileReader in SaveDictStep

hit-lacus:


engine-mr/src/main/java/org/apache/kylin/engine/mr/streaming/SaveDictStep.java
```diff
@@ -136,6 +136,7 @@ public boolean accept(Path path) {
                         logger.error("dictionary of column {} not found! ", colRef.getName());
                     }
                 }
+                reader.close();
```
shaofengshi: Will it be more safe to put it into a "finally" block?

hit-lacus: Thank you!

engine-mr/src/main/java/org/apache/kylin/engine/mr/streaming/SaveDictStep.java
```diff
@@ -108,32 +108,32 @@ public boolean accept(Path path) {
...
-                            new ByteArrayInputStream(dictInfo.getBytes())));
-
-                    Dictionary dict = dictionaryInfo.getDictionaryObject();
-                    if (dict != null) {
-                        dictionaryInfo = dictManager.trySaveNewDict(dict, dictionaryInfo);
-                        cubeSeg.putDictResPath(colRef, dictionaryInfo.getResourcePath());
-                        if (cubeSeg.getRowkeyStats() != null) {
-                            cubeSeg.getRowkeyStats().add(
-                                    new Object[] { colRef.getName(), dict.getSize(), dict.getSizeOfId() });
+                try(SequenceFile.Reader reader = new SequenceFile.Reader(fs, file.getPath(), conf)) {
```
shaofengshi: The code format here is not well. Please format it...

hit-lacus: Thank you, I have updated.

PR #1137, opened 2020-03-03
KYLIN-4400: use client in config to submit sql in system-cube.sh

zhoukangcn:
see: https://issues.apache.org/jira/browse/KYLIN-4400

use the same way in sample.sh


shaofengshi:
thank you Kang!

PR #1149, opened 2020-03-08
KYLIN-4399 Redirect http url is not override as https when using vip

sanjulian:


shaofengshi:
Hi Julian, will this configuration have any impact on existing (non-sso) users?

PR #1150, opened 2020-03-08
Kylin-4398 Display Kylin Version on UI

sanjulian:


nichunen:
Hi @sanjulian , thanks for your PR. 
In your pr, you add a rest api to get the version written as a static variable with name "org.apache.kylin.common.KylinVersion#CURRENT_KYLIN_VERSION". But we can not gurantee it is updated every time for the release, which means the version shown on UI may not be accurate.

Acutually, there is another api with code "org.apache.kylin.rest.controller.AdminController#getKylinVersions", I think the implementation there is more resonable.

sanjulian:
Please close the KYLIN-4398 pull request. It is duplicate with another ticket https://issues.apache.org/jira/browse/KYLIN-1850 which already release 3.0

sanjulian:
I have create another pull request for KYLIN-4399 https://github.com/apache/kylin/pull/1188

PR #1154, opened 2020-03-10
KYLIN-3847 Flink cubing step : fact distinct

harveyyue:
Refer jira: https://issues.apache.org/jira/browse/KYLIN-3847
Add flink fact distinct columns step, to avoid a mass of duplicated code with Spark and Flink implement self fact distinct step, add base class FactDistinctColumnsBase, extract core methods and variables from mapper/reducer class.  
About different engine will implement the abstract method FactDistinctColumnsBase.Visitor.collect(...) to output the result from Spark/Flink engine framework.
Please help to review this code changes


shaofengshi (approved):
+1 LGTM

PR #1163, opened 2020-03-17
KYLIN-4438 fix bug: null password may cause RuntimeException when starting up

xiacongling:
## Proposed changes
Add condition to treet null input for EncryptUtil methods.  For default settings, password may not be set, so `null` will passed to `PasswordPlaceholderConfigurer#resolvePlaceholder`.  This will cause an "IllegalArgumentException: Null input buffer" and lead to startup failure. This fix treets `null` as a legal input and returns `null` for both encrytion and decrytion.

## Types of changes

## Checklist

- [x] I have create an issue on [Kylin's jira](https://issues.apache.org/jira/browse/KYLIN-4438), and have described the bug/feature there in detail
- [x] Compiling and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] If this change need a document change, I will prepare another pr against the `document` branch
- [x] Any dependent changes have been merged

## Further comments
 N/A

### DIFF
--- a/core-common/src/main/java/org/apache/kylin/common/util/EncryptUtil.java
+++ b/core-common/src/main/java/org/apache/kylin/common/util/EncryptUtil.java
@@ -32,6 +32,9 @@ public class EncryptUtil {
             0x65, 0x79 };
 
     public static String encrypt(String strToEncrypt) {
+        if (strToEncrypt == null) {
+            return null;
+        }
         try {
             Cipher cipher = Cipher.getInstance("AES/ECB/PKCS5Padding");
             final SecretKeySpec secretKey = new SecretKeySpec(key, "AES");
@@ -45,6 +48,9 @@ public class EncryptUtil {
     }
 
     public static String decrypt(String strToDecrypt) {
+        if (strToDecrypt == null) {
+            return null;
+        }
         try {
             Cipher cipher = Cipher.getInstance("AES/ECB/PKCS5PADDING");
             final SecretKeySpec secretKey = new SecretKeySpec(key, "AES");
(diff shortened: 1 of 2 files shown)


core-common/src/main/java/org/apache/kylin/common/util/EncryptUtil.java
```diff
@@ -32,6 +32,9 @@
             0x65, 0x79 };
 
     public static String encrypt(String strToEncrypt) {
+        if (strToEncrypt == null) {
```
zzcclp: it's better use 'org.apache.commons.lang3.StringUtils.isBlank' to check blank string case.

xiacongling: Hi, zzcclp:
Thanks for the suggestion. But actually, a blank string can be encrypted and its cypher text is not a blank string. 
e.g. result of EncryptUtil.encrypt("") is "719bTBvphWewRiCiCj3D/A=="
So only null check is needed, I think.

zzcclp: Got it. Thanks for your explanation.

nichunen:
@xiacongling Hi, would you please open another issue for this pr, seems the change is not for KYLIN-4423

xiacongling:
Hi, @nichunen. Issue created for this PR, see [KYLIN-4438](https://issues.apache.org/jira/browse/)

nichunen:
@xiacongling OK, and the commit message should also start with the issue id, so we can link them on Jira

PR #1168, opened 2020-03-25
KYLIN-4433 use cube config when uhc step enabled

ChangjiGuo:
issue: https://issues.apache.org/jira/browse/KYLIN-4433
    If the system level configuration is used, it will cause the dictionary to be built repeatedly, so I think the configuration of the cube should be used here.
    Thanks!


shaofengshi:
Thank you Changji! Next time if you can add the JIRA id in git commit message, that would be great! like "KYLIN-4433 use cube config when uhc step enabled"

ChangjiGuo:
@shaofengshi ok, next time i will pay attention. I am very glad to contribute to the community!

PR #1214, opened 2020-05-19
[Bugfix] fix error with sending mail on ssl

chenqingspring:
## Proposed changes

Here is context of this problem：https://www.mail-archive.com/user@kylin.apache.org/msg04009.html

When I set `starttlsEnabled` to true, and specify a ssl port as 465,  the email could not be sent by this ssl port. It still use 25 default port. 

The root cause is, with such config above, HTMLEmail client doesn't enable ssl connection, so we need to enable ssl connection by `setSSLOnConnect` as well when using ssl port.

## Types of changes
[Bugfix] fix error with sending mail on ssl

### DIFF
--- a/core-common/src/main/java/org/apache/kylin/common/util/MailService.java
+++ b/core-common/src/main/java/org/apache/kylin/common/util/MailService.java
@@ -89,6 +89,7 @@ public class MailService {
         Email email = new HtmlEmail();
         email.setHostName(host);
         email.setStartTLSEnabled(starttlsEnabled);
+        email.setSSLOnConnect(starttlsEnabled);
         if (starttlsEnabled) {
             email.setSslSmtpPort(port);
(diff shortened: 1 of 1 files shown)


nichunen:
@chenqingspring Thanks for your pr, would you please open a jira for the issue?

chenqingspring:
@nichunen Sure, with my pleasure :)  [KYLIN-4515](https://issues.apache.org/jira/browse/KYLIN-4515)

chenqingspring:
@shaofengshi Hi xiaofeng, what do you mean by associating github commit? I'm trying to rename this issue start with KYLIN-4515, but it seems not editable.

chenqingspring:
@shaofengshi somehow ... this issue name could be changed. Tell me if I should do anything else?

PR #1245, opened 2020-06-08
KYLIN-4547 parse crc file error when building cube with mapreduce

qin7972:
## Proposed changes

Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.

## Types of changes

## Checklist

- [x] I have create an issue on [Kylin's jira](https://issues.apache.org/jira/browse/KYLIN), and have described the bug/feature there in detail

## Further comments

### DIFF
--- a/core-common/src/main/java/org/apache/kylin/common/persistence/ResourceStore.java
+++ b/core-common/src/main/java/org/apache/kylin/common/persistence/ResourceStore.java
@@ -682,6 +682,7 @@ abstract public class ResourceStore {
 
     public static class VisitFilter {
         public String pathPrefix = null;
+        public String pathExcludeSuffix = null;
         public long lastModStart = Long.MIN_VALUE;
         public long lastModEndExclusive = Long.MAX_VALUE;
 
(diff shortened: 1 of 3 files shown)


shaofengshi:
How does the crc file come? If it is not by kylin, then I think you need to find the root cause, instead of adding the filter.

qin7972:
> How does the crc file come? If it is not by kylin, then I think you need to find the root cause, instead of adding the filter.

qin7972:
@shaofengshi 
The crc file comes from hdfs.The path:
${yarn.nodemanager.local-dirs}/usercache/root/appcache/application_1591713180468_0013/container_1591713180468_0013_01_000001/meta

This error does not appear in kylin 2.5.1. 
Because kylin's CachedCrudAssist loads resources from the directory, it uses the json suffix rule to match the file, the file with the crc suffix is filtered out.
Then the JsonSerializer is used to deserialize the files that meet the conditions, so there will be no problems.

However, in kylin 3.0.0, CachedCrudAssis first scans all files in the directory, and deserializes at the same time. 
The file with the suffix crc is deserialized by JsonSerializer to report this error. 
After the scan is complete, use the json suffix rule to match the file name.

Our pr refer to kylin 2.5.1, which is to add Filter to sink and filter out the files with crc suffix.

LingangJiang:
How about this issue?  Although mapreduce job is success, but error in log

LingangJiang:
this issue also appear in flink job engine

PR #1251, opened 2020-06-10
KYLIN-4548 fix garbled characters in sql when exporting query result

ChangjiGuo:
when the SQL that needs to export the query results has Chinese, garbled characters will be generated.
jira: https://issues.apache.org/jira/browse/KYLIN-4548


ChangjiGuo:
Can you take time to review this pr? @shaofengshi

PR #1292, opened 2020-06-29
KYLIN-4607 destributed scheduler kill job instead of  fetcherRunner k…

bigxiaochu:


bigxiaochu:
@RupengWang how do you see.

PR #1299, opened 2020-06-30
KYLIN-4613 add build cubev2 and JobRestClient

bigxiaochu:


RupengWang:
+ 1
Test Passed with the following command:
```shell
./bin/kylin.sh org.apache.kylin.tool.job.CubeBuildingCLIV2 --cube kylin_sales_cube --endTime 1338508800000 --host localhost --port 7971 --userName ADMIN --password KYLIN
```

bigxiaochu:
@xiaowangyu how do you see.

xiaowangyu:
> @xiaowangyu how do you see.

I think the change of JobRestClient and CubeBuildingCLIV2 can be update in original class may be better. Compatible with the original implementation. For users, the interface is unified.

bigxiaochu:
@xiaowangyu  do you mean change of CubeBuilding? 
It will be umcompatibility . I think compatibility is import, so use v2

core-common/src/main/java/org/apache/kylin/common/KylinConfig.java
```diff
@@ -134,7 +134,7 @@ private static void buildDefaultOrderedProperties() {
         }
     }
 
-    public static KylinConfig getInstanceFromEnv() {
+    public static KylinConfig getInstanceFromEnv(boolean allowConfigFileNoExist) {
```
hit-lacus, line 137: Why we need `allowConfigFileNoExist` ?

bigxiaochu: 因为作为客户端提交建模、构建请求时，可以不需要有kylin的部署结构，也不需要有符合部署目录结构规范的配置文件。以减少客户端程序的依赖复杂度。

hit-lacus (approved):
Approved

hit-lacus:
This patch looks good to me, thanks @bigxiaochu . We should consider to write documentation for this tools.

PR #1311, opened 2020-07-09
KYLIN-4626 add set kylin home shell

bigxiaochu:


build/bin/set-kylin-home.sh
```diff
@@ -0,0 +1 @@
+export KYLIN_HOME=$KYLIN_HOME
```
shaofengshi, line 1: hi xiao, what's the behavior of this line? If KYLIN_HOME already been set, will export it again bring any change?

bigxiaochu:
@xiaowangyu how do you see

hit-lacus (approved):
Approved, thank you @bigxiaochu .

PR #1351, opened 2020-08-06
KYLIN-4683 Fail to consume kafka when partition number get larger

tianhui5:
## Proposed changes

JIRA：https://issues.apache.org/jira/browse/KYLIN-4683

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update (if none of the other choices apply)

## Checklist

- [x] I have create an issue on [Kylin's jira](https://issues.apache.org/jira/browse/KYLIN), and have described the bug/feature there in detail
- [ ] Commit messages in my PR start with the related jira ID, like "KYLIN-0000 Make Kylin project open-source"
- [ ] Compiling and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] If this change need a document change, I will prepare another pr against the `document` branch
- [ ] Any dependent changes have been merged

PR #1392, opened 2020-09-03
KYLIN-4743 NPE when some kafka partition's offset is not exists in checkpoint

ggKe:
https://issues.apache.org/jira/projects/KYLIN/issues/KYLIN-4743?filter=allopenissues


hit-lacus:
Thank you @ggKe , looks like this patch has been fixed by https://github.com/apache/kylin/pull/1351 , so thank you for your contribution.

PR #1415, opened 2020-09-16
Error while compiling statement: FAILED: SemanticException Line 1:21 Table or database name may not contain dot(.) character

lijianfeng6666526:
The version is :
kylin2.6.3  + hadoop3.1

**when the job cube , an error has occurred . the error is as follows** **:** 

image

image

image


hit-lacus:
I will close this, you can reopen it if you need.

PR #1462, opened 2020-10-24
KYLIN-4736 flink upgrade to 1.11.1

chenjie-sau:
## Proposed changes

flink upgrade to 1.11.1

### DIFF
--- a/core-common/src/main/java/org/apache/kylin/common/util/HadoopUtil.java
+++ b/core-common/src/main/java/org/apache/kylin/common/util/HadoopUtil.java
@@ -64,12 +64,14 @@ public class HadoopUtil {
             return conf;
         }
         Configuration conf = hadoopConfig.get();
+        conf.set("fs.hdfs.impl.disable.cache", "true");
         return conf;
     }
 
     public static Configuration healSickConfig(Configuration conf) {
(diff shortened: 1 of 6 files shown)


chenjie-sau:
@yanghua could you please take a look when free, thank you

yanghua (approved):
LGTM. @zhangayqian Would you please help to review this new PR?

zhangayqian:
@chenjie-sau @yanghua Thanks！LGTM，works well in my local env (cdh57).

core-common/src/main/java/org/apache/kylin/common/util/HadoopUtil.java
```diff
@@ -64,12 +64,14 @@ public static Configuration getCurrentConfiguration() {
             return conf;
         }
         Configuration conf = hadoopConfig.get();
+        conf.set("fs.hdfs.impl.disable.cache", "true");
```
zzcclp, line 67: Why it needs to add this parameter here? and will it impact other functions, because there are some place calling this functions?

chenjie-sau: This configuration will not affect other functions, it has been verified,This is done in order to prevent the FileSystem from shutting down prematurely, resulting in abnormal file information acquisition

PR #1534, opened 2021-01-02
KYLIN-4855 kylin.metrics.prefix bug fix

wolfboys:
KYLIN-4855 kylin.metrics.prefix bug fix

## Proposed changes

After executed "system-cube.sh cron", the system cube will be added to the crontab, but the cube name is wrong, and it does not match the actual situation. After testing, it is found that the system-cube is
There is a bug in the handling of kylin metrics prefix

In the system-cube.sh script, the prefix is "KYLIN". The logic is wrong. You should get "kylin.metrics.prefix" from kylin.properties. If this item is set, take the value set by the user, if not set, go to the default value. "KYLIN"

## Types of changes

## Checklist

- [x] I have create an issue on [Kylin's jira](https://issues.apache.org/jira/browse/KYLIN-4855), and have described the bug/feature there in detail
- [ ] Commit messages in my PR start with the related jira ID, like "KYLIN-0000 Make Kylin project open-source"

## Further comments


wolfboys:
Update as follows:


1) Import bin in unified script/ header.sh Code style, all using source ${KYLIN_HOME:-"$(cd -P -- "$(dirname -- "$0")" && pwd -P)/../"}/bin/header.sh

2) There are code format problems in unified scripts. 1 tab and 2 tab can be seen everywhere. Now the unified code is indented to 2 tab

3)  cached-*- dependency.sh Change to .cached-*- dependency.sh The file is not directly visible by default

4) Underutilized cached-*- dependency.sh File. Only some files use the dependent cache file. Now we use the dependent cache file globally to increase the startup time after the first time

5) Repair incremental build- cube.sh The problem of building user name and password in the system (if the admin password changed cannot be work)

6) Fix multiple execution of "system- cube.sh Cron " will repeatedly add tasks to crontab

PR #1549, opened 2021-01-18
KYLIN-4879 Fix removeCommentInSql function

wangjie-fourth:
## Proposed changes

recode removeCommentInSql method of QueryUtil. In some case, the original code can not process right.
you can see detail in [KYLIN-4879](https://issues.apache.org/jira/browse/KYLIN-4879)

If you think something is not good, please comment below, i will keep eye on it.

## Types of changes

## Checklist

- [x] I have create an issue on [KYLIN-4879](https://issues.apache.org/jira/browse/KYLIN-4879), and have described the bug/feature there in detail
- [x] Compiling and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] If this change need a document change, I will prepare another pr against the `document` branch

## Further comments


wangjie-fourth:
@hit-lacus hi,could you please help me ?`LGTM analysis: Java` seems to fail because it did not download the jar package successful, Should I re-push to trigger these checks?

hit-lacus:
Hi @wangjie-fourth , thanks for your contribution, I have re-trigger a re-run for **LGTM analysis: Java**, let's wait and see.

wangjie-fourth:
@hit-lacus hi,it is done already. could you please review ?

hit-lacus:
@wangjie-fourth  Sorry for late reply, the patch looks good to me, but I wonder if it is OK that you can use Apache Calcite API to do the same things. 

From my oponion, deleting comments is a relative trivial matter, it is not worthwhile to introduce an external(not from apache but using apache license) dependency.

wangjie-fourth:
@hit-lacus you are right.I will use calcite to solve this problem later

hit-lacus:
Thanks @wangjie-fourth , you are great .

wangjie-fourth:
@hit-lacus Hi, could you please review ?

wangjie-fourth:
@hit-lacus Hi, when I try to use Apache Calcite to solve this problem, I realized something

The function of this method includes:
(1) Correctly delete comments in sql
(2) The parameter string may contain multiple SQL,and could be DDL、DML...
(3) The result of string should retain the original sql style;E.g, select need be select,can not converted to SELECT
(4) This function does not need to verify whether the SQL conforms to the SQL syntax specification

I can't find a suitable API in Apache Calcite to meet the above requirements.In my opinion, even using Apache Calcite's API to solve this problem will be very troublesome.
Maybe it’s because I’m not very familiar with Apache Calcite. If you have a good solution,please share it with me.

So,I wrote a SqlCommentParser.class to solve this problem.Do you think this solution meets your requirements?

hit-lacus:
Thanks a lot for taking a lot of effort on this issues, I was thinking if it is a better to solve this via javacc. I am doing a small research on this. Will update soon.

PR #1615, opened 2021-03-19
[KYLIN-4939] Transform lookup table snapshot from segment level to cu…

Ted-Jiang:
…be level

## Proposed changes

We find many use cases should use cube-level lookup table snapshot instead of segment-level ones. For segment-level ones, it will cause many issues. For example for table BATCH_VIEWS.xxx, it changes everyday. Now there're around 2000 snapshots.

For building a new snapshot, it will check the duplication to avoid duplicated snapshots, which means it will check all of the 2000 snapshots. It really costs time.
For query & building, to cache the snapshot, if each takes around 6MB, the total will need around 12GB memory.
After this change, we can not only speedup cubing but also reduce the burden of memory cost.

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Checklist

- [ ] I have create an issue on [Kylin's jira](https://issues.apache.org/jira/browse/KYLIN), and have described the bug/feature there in detail
- [ ] Commit messages in my PR start with the related jira ID, like "KYLIN-0000 Make Kylin project open-source"

## Further comments

### DIFF
--- a/core-cube/src/main/java/org/apache/kylin/cube/CubeDescManager.java
+++ b/core-cube/src/main/java/org/apache/kylin/cube/CubeDescManager.java
@@ -256,6 +256,18 @@ public class CubeDescManager {
         }
     }
 
+    public CubeDesc updatelookupTableSnapshotGlobal(CubeDesc desc, String lookupTable, boolean global) throws IOException {
+        try (AutoLock lock = descMapLock.lockForWrite()) {
+            CubeDesc copy = desc.latestCopyForWrite();
+            copy.createAndSetSnapshotTableGlobal(lookupTable, global);
+            return updateCubeDesc(copy);
+        }
+    }
+
+    public CubeDesc copyForWrite(CubeDesc desc) {
+        return crud.copyForWrite(desc);
+    }
+
     /**
      * if there is some change need be applied after getting a cubeDesc from front-end, do it here
      * @param cubeDesc
--- a/core-cube/src/main/java/org/apache/kylin/cube/model/CubeDesc.java
+++ b/core-cube/src/main/java/org/apache/kylin/cube/model/CubeDesc.java
@@ -59,6 +59,7 @@ import org.apache.kylin.common.persistence.RootPersistentEntity;
 import org.apache.kylin.common.util.Array;
 import org.apache.kylin.common.util.JsonUtil;
 import org.apache.kylin.common.util.Pair;
+import org.apache.kylin.cube.CubeDescManager;
 import org.apache.kylin.cube.cuboid.CuboidScheduler;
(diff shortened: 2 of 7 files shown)


core-cube/src/main/java/org/apache/kylin/cube/CubeDescManager.java
```diff
@@ -256,6 +256,18 @@ public CubeDesc updateCubeDesc(CubeDesc desc) throws IOException {
         }
     }
 
+    public CubeDesc updateCubeDescToBeGlobal(CubeDesc desc, String lookupTable, boolean global) throws IOException {
```
hit-lacus, line 259: The name of this method is misleading.

core-cube/src/main/java/org/apache/kylin/cube/model/CubeDesc.java
```diff
@@ -623,6 +624,12 @@ public boolean consistentWith(CubeDesc another) {
         return this.calculateSignature().equals(another.calculateSignature());
     }
 
+    public CubeDesc lastedCopyForWrite() {
```
hit-lacus, line 627: Typo, "lasted" or "latest" ?

server-base/src/main/java/org/apache/kylin/rest/controller/CubeController.java
```diff
@@ -346,6 +346,35 @@ public JobInstance rebuildLookupSnapshot(@PathVariable String cubeName,
...
+        List<CubeInstance> result = new ArrayList<>();
+
+        final CubeManager cubeMgr = cubeService.getCubeManager();
+        String[] changeCubes = cubeNames.toUpperCase(Locale.ROOT).split(",");
+        for (String cubeName : changeCubes) {
+            try {
+                checkCubeExists(cubeName);
+                final CubeInstance cube = cubeMgr.getCube(cubeName);
+                CubeInstance cubeInstance = cubeService.changeLookupSnapshotBeGlobal(cube, tableName);
+                logger.info("cube %s change snapshotTable %s global Success", cubeName, tableName);
```
hit-lacus, line 368: Placeholder should be `%` or `{}` ?

Ted-Jiang: fix

core-cube/src/main/java/org/apache/kylin/cube/model/CubeDesc.java
Ted-Jiang: thx!

hit-lacus (approved):
Approved

PR #1622, opened 2021-04-08
[KYLIN-4964] Receiver consumer thread should be stoped while encounting unrecoverable error

dixingxing0:
## Proposed changes

org.apache.kylin.stream.core.storage.columnar.ColumnarMemoryStorePersister#persist will catch exception and just log it, this behavior will cause critical problem, like our case: there is no space left on device, `persist` method failed for many times, we lost several hours of data.

## Types of changes

- [x ] Bugfix (non-breaking change which fixes an issue)

## Checklist

- [ x] I have create an issue on [Kylin's jira](https://issues.apache.org/jira/browse/KYLIN), and have described the bug/feature there in detail
- [ x] Commit messages in my PR start with the related jira ID, like "KYLIN-0000 Make Kylin project open-source"
- [ x] Compiling and unit tests pass locally with my changes

## Further comments

Here is our solution which already tested:

1.Throw IllegalStorageException while we cannot do the persist
2.Then stop the consumer thread
3.Add `consumer_thread_alive` in ConsumerStats for monitoring
4.Also fix another issue which will cause wrong result after restore from checkpoint


hit-lacus:
LGTM, thx @dixingxing0

PR #1626, opened 2021-04-11
[KYLIN-4554] validate filter condition on model saving

zheniantoushipashi:
verify filter condition using the following login

1 select * from tableName where {filterCondition} pass   calcite  parse

2 all column in filter condition must be model table column

### DIFF
--- a/core-job/src/main/java/org/apache/kylin/job/JoinedFormatter.java
+++ b/core-job/src/main/java/org/apache/kylin/job/JoinedFormatter.java
@@ -56,6 +56,14 @@ public class JoinedFormatter {
         setDateEnv(flatDesc);
     }
 
+    public JoinedFormatter(Boolean validateModel) {
+        // for validate model filter condition
+        String start = "20190710";
+        String end = "20190711";
+        setKeyValue(START_DATE, start);
+        setKeyValue(END_DATE, end);
+    }
+
     private void setDateEnv(IJoinedFlatTableDesc flatDesc) {
         DataModelDesc model = flatDesc.getDataModel();
         PartitionDesc partDesc = model.getPartitionDesc();
@@ -83,7 +91,7 @@ public class JoinedFormatter {
         return value == null ? "" : value;
     }
 
-    String formatSentence(String sentence) {
+    public String formatSentence(String sentence) {
         String[] cArray = REG_PATTERN.split(sentence);
         StringBuilder sbr = new StringBuilder();
         List<String> keys = getKeys(sentence);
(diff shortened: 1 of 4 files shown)


hit-lacus:
Did you patch consider dynamic variable case?


image

zheniantoushipashi:
**test evidence** 


**Positive testing**

Use ssb data set， create a  data model,    add filter LO_CUSTKEY > 1  at  last step 

Successfully created the model

image


**Negative Testing** 

Add an additional Y after the LO_CUSTKEY column to construct a column ：LO_CUSTKEYY   that does not belong to the model table

image

The following error is reported when saving the model

Failed to deal with the request: java.lang.IllegalArgumentException: filter condition col: LO_CUSTKEYY is not a column in the table

Meet the test expectations

zheniantoushipashi:
> Did you patch consider dynamic variable case? Like `date_str>='${start_date}' and date_str<'${end_date}'` .
> 
> image

thanks for reminding，  When saving the model, use the start_date and end_date of a mock to replace sql, and then verify

see the below test evidence

image

hit-lacus:
Maybe we can make it more Robust?

I change your UT a little, by changing from `TRANS_ID` to `kylin_account.TRANS_ID`, and find a mistake.

```java
deserialize.setFilterCondition("kylin_account.TRANS_ID = 1");
```

Following is the debug mode, it indicates that `allSqlIdentifier.add(id.names.get(0));` return `kylin_account`, not column name.

image

I think it could be better implemented, am I right ?

hit-lacus:
Besides, I think columns in dimension table should also be taken into account.

hit-lacus (changes requested):
-1
I think more cases should be considered. Do you agreed?

zheniantoushipashi:
retest this please 

> Maybe we can make it more Robust?
> 
> I change your UT a little, by changing from `TRANS_ID` to `kylin_account.TRANS_ID`, and find a mistake.
> 
> ```java
> deserialize.setFilterCondition("kylin_account.TRANS_ID = 1");
> ```
> 
> Following is the debug mode, it indicates that `allSqlIdentifier.add(id.names.get(0));` return `kylin_account`, not column name.
> 
> image
> 
> I think it could be better implemented, am I right ?

yes, you are right, should consider it

zheniantoushipashi:
I added all the columns of all tables to the validate collection,  I did verification in ut and integration tests,

**test evidence**

image


image

hit-lacus (approved):
LGTM

hit-lacus:
Since that the version 3.1.2 is going to releasing, I would merge any new code change  at the moment. I will merge it after 3.1.2 is released. 

Thanks for your first contribution, @zheniantoushipashi please go ahead!

zheniantoushipashi:
retest this please

PR #1628, opened 2021-04-12
[KYLIN-4554] validate filter condition on model saving

zheniantoushipashi:
verify filter condition using the following login

1 select * from tableName where {filterCondition} pass   calcite  parse

2 all column in filter condition must be model table column

### DIFF
--- a/core-metadata/src/main/java/org/apache/kylin/metadata/util/ModelUtil.java
+++ b/core-metadata/src/main/java/org/apache/kylin/metadata/util/ModelUtil.java
@@ -1,0 +1,103 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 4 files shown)


core-metadata/src/main/java/org/apache/kylin/metadata/util/ModelUtil.java
```diff
@@ -0,0 +1,103 @@
...
+            });
+        }
+
+        public Object visit(SqlIdentifier id) {
+            allSqlIdentifier.add(id.names.get(0));
+            return null;
+        }
+    }
+
+    public static String doubleQuotaKeywordDefault(String sql) {
```
hit-lacus, line 96: `Quota` or `Quote` ?

zheniantoushipashi: got

hit-lacus (approved):
LGTM

PR #1821, opened 2022-02-24
I18n

zhenfuxu:
webapp i18n 我完成了国际化，(KYLIN-5109) webapp i18n, I will maintain the Chinese localization.


zhenfuxu:
i will do some others things.

PR #2162, opened 2023-12-26
KYLIN-5748 Skip build flat table for index build job if all new created indexes can be built from existed parent index

zznlime:
For an index build job, if all the new added indexes can be built from their existed parent index. No index was built from new created flat table.
For this case, there won't be any data inconsistency across indexes. And there is no need to build a new flat table to serve the stage of index build by layer.
So I think we can skip stage BUILD DICT and stage GENERATE FLAT TABLE for an index build job in condition that all new added indexes can be built from their existed parent index

### DIFF
--- a/src/core-metadata/src/main/java/org/apache/kylin/metadata/cube/model/NBatchConstants.java
+++ b/src/core-metadata/src/main/java/org/apache/kylin/metadata/cube/model/NBatchConstants.java
@@ -42,6 +42,7 @@ public interface NBatchConstants {
     String P_PARTITION_IDS = "partitionIds";
     String P_BUCKETS = "buckets";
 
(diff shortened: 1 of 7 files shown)


zznlime:
@hit-lacus PTAL, Thanks

Mrhs121:
Your proposal will be invalid in this use case: if you add a new bitmap type metric to the model, you can build the new index separately, and the task type is also INDEX_BUIDL. If you skip building the dictionary, an error will occur when generating encoding columns for the flat table later.

pfzhan:
@zznlime Does this pr need to be merged into the kylin code? If necessary, please send pr to @liyang-kylin . I will resolve this issue.

PR #2164, opened 2024-01-24
fix multi session password change

wuzhim:
Clear the authentication cache to prevent logging in with the old password after a password reset.
Expire the user's related session after their password reset.


wuzhim:
I had run IT use `mvn test -DskipRunIt=false` in my local env. @hit-lacus

PR #2251, opened 2025-02-19
KYLIN-5997 fix query with streaming model

gleonSun:


jlfsdtc:
The CI execution result of this PR is shown in the figure:
image

PR #2304, opened 2025-02-27
KYLIN-6055 Support for calculating dimension_range_information_map for one level partition under multi partitions

gleonSun:


lionelcao (approved):
LGTM

PR #2310, opened 2025-02-27
KYLIN-6061 build enable gluten

gleonSun:
1. Enable Gulten for Async query
2. Enable Gulten for Build Job
3. fix build stage execution progress exception


lionelcao (approved):
LGTM

PR #2315, opened 2025-02-27
KYLIN-6066 Adapt the spark-test script for gluten and opt code structure

gleonSun:
* support cluster for spark-test with enable kerberos

* Add default gluten.jar classpath for sparder driver

* fix get yarn cluster info


lionelcao (approved):
LGTM

PR #2316, opened 2025-02-27
KYLIN-6067 Keep columns order in select star for ModelView

gleonSun:
* change order config from int to string

* fix toLowerCase


lionelcao (approved):
LGTM

